PR #2, opened 2017-01-07
test CSG

dav-m85:
Testing the CSG shape generation, first attempt.


test.js
```diff
@@ -0,0 +1,137 @@
...
+            {_x:0,_y:0,_z:0},
+            {_x:1,_y:1,_z:1},
+            {_x:-1,_y:1,_z:1}
+        ]},
+        b = {vertices:[
+            {_x:-1,_y:1,_z:1},
+            {_x:0,_y:0,_z:0},
+            {_x:1,_y:1,_z:1}
+        ]};
+    t.true(comparePolygons(a, b));
```
kaosat-dev: This might just be me, but is the order of vertices not significant ? Ie should these actually not be unequal ?
Usually it impacts winding order for polygonal models so could lead to issues.

kaosat-dev:
Thanks for getting started with this  ! A big step in the right direction !
I agree with the need to add helpers to compare polygons etc, manual compare like I did for some tests can get tedious.

test.js
```diff
@@ -0,0 +1,137 @@
...
+            {_x:0,_y:0,_z:0},
+            {_x:-1,_y:1,_z:1},
+            {_x:1,_y:1,_z:1}
+        ]};
+    t.false(comparePolygons(a, b));
+});
+
+// Constructive operations
+// a + b = c
+let a = CSG.cube({
```
kaosat-dev: Not entirely sure that creating CSG 'shapes' outside out the scope of tests will work correctly: 
all tests are run in parallel in AVA , so it might lead to weird side effects.

Perhaps just refactor it like this, (bonus: it is also a test helper function)
```javascript
function createOperands(){
const a = CSG.cube({
     center: [0,0,0],
     radius: [1,1,1]
 });
const b = CSG.cube({
     center: [2,0,0],
     radius: [1,1,1]
 });
 const c = CSG.cube({
     center: [1,0,0],
     radius: [2,1,1]
 })
 return {a, b, c}
}
```
then you can just use them in each test :
```javascript 
test('CSG.union', t => {
    const {a, b, c} = createOperands()
     sameCSG(a.union(b), c) ? t.pass() : t.fail();
 });

```

test.js
dav-m85: Let me illustrate with a drawing:
image001

Polygons ABCD (black) and ABDC (red) share the same vertex, but not the same edges. I consider them different. comparePolygons is however returning true if the vertex are messed up with but give the same edges (like BCDA or CDAB, that can be rendered as the black polygon here).

kaosat-dev: Ok thanks for clarifying that ! :)
I think it is ok to check it in this manner  :  
> 'purpose here is to check that the solid created has the right shape'
that seems good enough for now , we can always flesh out down the line .

z3dev:
@dav-m85 Thanks for getting this started.

I have some tests as well for CSG and CAG. But these need to be reworked to use AVA. I'm using CHAI at the moment.

cag-common-transforms.js
cag-conversions.js
cag-new.js
cag-shapes.js
csg-new.js

Just one comment, this library will expand rapidly over the next few months. So, let's name the tests appropriately so others can see what's already completed by the names.

test.js
z3dev: Nice tip. I'll keep that in mind when testing more complex functions and geometries.

kaosat-dev:
hi @dav-m85  , sorry to bother you , but can we perhaps help you with the tests ? @z3dev added a few, but yours are also very welcome !
Cheers !

z3dev:
@dav-m85 Can you make changes to make the merge seamless? We want to keep the momentum going on testing.

dav-m85:
Sure ! On it !

Le 17 févr. 2017 12:22 PM, "Z3 Development" <notifications@github.com> a
écrit :

csg.js
```diff
@@ -1503,7 +1501,7 @@ for solid CAD anyway.
      * });
      */
     CSG.cube = function(options) {
-        var c, r;
+        var c, r, corner1, corner2;
```
dav-m85: cube test was failing because of those missing var declarations

helpers/obj-store.js
```diff
@@ -1,71 +1,45 @@
...
+};
+
+OBJ.save = function(objectid,object) {
+  fs.writeFileSync(_path(objectid), JSON.stringify(object), 'utf8');
+};
+
+OBJ.load = function(objectid) {
+  var buffer = fs.readFileSync(_path(objectid),'utf8');
+  var bin = JSON.parse(buffer);
+  if ("sides" in bin) {
```
dav-m85: you can easily infer the binary type by looking for "sides" or "polygons" member.

package.json
```diff
@@ -52,10 +52,6 @@
   "dependencies": {},
   "devDependencies": {
     "ava": "^0.15.2",
-    "babel": "^6.5.2",
```
dav-m85: as mentionned in https://github.com/jscad/csg.js/issues/14, those deps are actually already imported by ava

test/cag-common-transforms.js
```diff
@@ -61,18 +61,7 @@ test('CAG should intersect properly', t => {
...
-})
-
-test('CAG should expand properly', t => {
-  t.fail('TODO: add tests') // non implemented tests should fail so we see the changes when we actually add them!!
-})
-
-test('CAG should contract properly', t => {
-  t.fail('TODO: add tests') // non implemented tests should fail so we see the changes when we actually add them!!
-})
+test.todo('CAG should transform properly');
```
dav-m85: @z3dev I personally prefer to flag the todos with this ava suffix, hope you don't mind.

test/csg-conversions.js
```diff
@@ -0,0 +1,16 @@
+import test from 'ava';
+import {CSG} from '../csg';
+import {OBJ} from '../helpers/obj-store';
+import {assertSameGeometry} from '../helpers/asserts';
+
+test("CSG can be packed into and retrieved from a compact binary", t => {
```
dav-m85: Maybe we should add documentation to explain that compactBinary cannot be JSONified, for reasons I don't know (UInt32Array issue ?)

dav-m85:
@kaosat-dev @z3dev I think I'm done for this first attempt.

dav-m85:
meh... test fails only on node v4.8.0 :(

test/cag-common-transforms.js
kaosat-dev: agreed, this makes it much shorter & clearer !

csg.js
kaosat-dev: nice catch !

kaosat-dev:
@dav-m85 hmm weird, the babel preset file is present ? not sure why it is complaining

z3dev:
Excellent additions. Thanks

package.json
```diff
@@ -52,10 +52,6 @@
...
-    "babel-cli": "^6.18.0",
-    "babel-core": "^6.21.0",
-    "babel-preset-es2015": "^6.18.0",
```
kaosat-dev: @dav-m85 these are why the tests fails in node 4.x , sorry , need to add them back, they need to be explictely added

dav-m85:
@kaosat-dev @z3dev test passing ;)

kaosat-dev:
woohoo ! well done ! awesome @dav-m85  !
Merging ! :)

PR #3, opened 2017-01-15
Added tests for new CSG and new CAG objects

z3dev:
Please review, as I'd like these tests committed before submitting another PR for changes to csg.js


kaosat-dev:
Looks good to me so far, very in dept testing.
It seems like the change of paths (test vs tests) did not get handled correctly though as there are 4 files in the PR but some are identical except for the path ?

z3dev:
It seems that when moving, you have to commit both deleted and added files. I will commit the delete... SIGH

Thanks for the review.

kaosat-dev:
Nice to see all the new tests coming in , awesome work !

z3dev:
A good start, so please review and comment. If these are fine then please merge.

kaosat-dev:
Looked over the various tests, and it looks like a great start indeed!

PR #6, opened 2017-01-25
Small fixes from testing

z3dev:
Just a few small fixes from testing various functions of CSG and CAG.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ PRs accepted.
 
 Small note: If editing the Readme, please conform to the [standard-readme](https://github.com/RichardLitt/standard-readme) specification.
 
+## Copyrights
+
+Some copyrights apply. Copyright (c) 2012 Joost Nieuwenhuijse (joost@newhouse.nl), under the MIT license. Copyright (c) 2011 Evan Wallace (http://madebyevan.com/)
 
 ## License
 
--- a/csg.js
+++ b/csg.js
@@ -1983,11 +1983,11 @@ for solid CAD anyway.
                                 this._z = 0;
                             }
                         }
-                    } else if (('_x' in x) && ('_y' in x)) {
-                        this._x = parseFloat(x._x);
-                        this._y = parseFloat(x._y);
-                        if ('_z' in x) {
-                            this._z = parseFloat(x._z);
+                    } else if (('x' in x) && ('y' in x)) {
+                        this._x = parseFloat(x.x);
+                        this._y = parseFloat(x.y);
+                        if ('z' in x) {
+                            this._z = parseFloat(x.z);
                         } else {
                             this._z = 0;
                         }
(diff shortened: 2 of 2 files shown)


csg.js
```diff
@@ -5483,17 +5483,18 @@ for solid CAD anyway.
     // Each side is a line between 2 points
     var CAG = function() {
         this.sides = [];
-        this.isCanonicalized = false;
+        this.isCanonicalized = true;
```
kaosat-dev: That seems like a very significant change , why does the default need to be changed ? It at least impacts a portion of SCAD-API tests and potentially the behaviour of the library

csg.js
```diff
@@ -1983,11 +1983,11 @@ for solid CAD anyway.
                                 this._z = 0;
                             }
                         }
-                    } else if (('_x' in x) && ('_y' in x)) {
-                        this._x = parseFloat(x._x);
-                        this._y = parseFloat(x._y);
-                        if ('_z' in x) {
-                            this._z = parseFloat(x._z);
+                    } else if (('x' in x) && ('y' in x)) {
```
kaosat-dev: I am not entirely sure about this one : does this mean that the input object should have fields x & y instead of _x & _y ?

README.md
```diff
@@ -40,6 +40,9 @@ PRs accepted.
 
 Small note: If editing the Readme, please conform to the [standard-readme](https://github.com/RichardLitt/standard-readme) specification.
 
+## Copyrights
```
kaosat-dev: This is good thanks ! 
But given that the other changes are less obvious perhaps it would be best to seperate this into another PR please :)

csg.js
z3dev: These fields names changes are necessary to allow CAG to be stored as anonymous objects, and then restored from those via CAG.fromObject(). Basically, this fixes CAG.fromObject().

csg.js
z3dev: First, this makes the CAG constructor consistent the CSG constructor. Second, this change makes effects the speed of operations performed with a newly created CAG, i.e. Canonization is not required.

I haven't seen any side effects. And it may be possible to remove this "state" with proper mathematics later.

z3dev:
I need to revisit these changes again. It seems that I'm not consistent on changes here and the previous repository.

csg.js
kaosat-dev: Ok double checked in the JSCAD-SCAD-API code, should be good !

csg.js
kaosat-dev: Ok , do you mean this one ?
https://github.com/z3dev/csg.js/blob/207c06310244937436d8c5f3c088c7ac457ace80/csg.js#L6422 ? so this fix would allow passing the object directly, without the need for obj.pos._x,obj.pos._y ?

PR #7, opened 2017-01-28
Integrating Changes from Original gh-pages

z3dev:
And added test cases too.


kaosat-dev:
This is great ! Thanks for integrating these ! I see the tests all pass too !

PR #9, opened 2017-02-04
Update for Automated Testing as well as CSG

z3dev:
These changes will allow the automated testing to complete properly.

There's also a start at testing CSG shapes.


z3dev:
FYI, the results from Travis look correct at this point in testing. 7 failed tests.

PR #28, opened 2017-03-05
Completed integration of cylinderElliptic()

z3dev:
small set of changes to complete the integration of this new functionality

sorry, no tests. just a stub.


dav-m85:
Good pick. Maybe we should think about implementing a code style validator ?

kaosat-dev:
Even if it is not preferable in the future to not add tests, this looks good, so are you guys ok if I merge ?

z3dev:
If we require tests with each PR then we should state specifically. Anyway, this is a good addition so let's merge.

PR #30, opened 2017-03-25
Initial Tests for Vector2D

z3dev:
Let's keep those tests rolling in!!!!

### DIFF
--- a/helpers/nearlyEqual.js
+++ b/helpers/nearlyEqual.js
@@ -1,0 +1,28 @@
+// Compare two numeric values for near equality.
+// the given test is fails if the numeric values are outside the given epsilon
+function nearlyEqual(t, a, b, epsilon, failMessage) {
+  if (a == b) { // shortcut, also handles infinities and NaNs
+    return true;
+  }
+
+  var absA = Math.abs(a);
+  var absB = Math.abs(b);
+  var diff = Math.abs(a - b);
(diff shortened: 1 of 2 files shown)


kaosat-dev:
Great set of tests @z3dev !
Only one issue : failing Travis build due to a single rounding error:  is that test passing for you locally ?

z3dev:
All test pass on Mac OS. I was wondering if there would be issues with rounding. We need another test for these values, something like "equal within 0.000001"

kaosat-dev:
Since ava does not support other assertions (https://github.com/avajs/ava/issues/1094)

The simplest solution is this one 
```javascript
const precision = 0.000001
t.deepEqual(Math.abs(expected-observed) < precision) 
```
and voila ! solved :)
simple and straight to the point I think
if we really want to we could wrap it for re-use

```javascript
function almostEquals(t, observed, expected, precision=0.0000001){
t.deepEqual(Math.abs(expected-observed) < precision) 
}
```

and use it like this
```javascript
almostEquals(t, observed, expected)
```

z3dev:
Looks like a function is required, which is explained here... http://floating-point-gui.de/errors/comparison/

I'll add a helper for this.

z3dev:
@kaosat-dev please review the helper function. i tested the different cases of comparisons. Seems to work fine.

kaosat-dev:
Yup , helper is ok, and all tests are passing, well done !

Issue #29, opened 2017-03-19
Tolerances


z3dev:
@joostn I'm revisiting all the hard coded tolerances in the library, and trying to make sense out of the values being applied.

I've introduced CSG.ESP = 1e-5 for the default value, but think there should be a few more tolerances available. Maybe...

CSG.distanceEPS = 1e-5
CSG.areaEPS = (CSG.distanceEPS*CSG.distanceEPS)/2;
CSG.angleEPS = 1e-3; // percent of angle

Thoughts?

joostn:
What would you need those other tolerances for?
If i recall correctly the CSG operations need a tolarance value to decide whether a point is on a plane or not. (each CSG operation creates small errors due to rounding). So that would be just a 1D distance value.

z3dev:
First, I would like to remove the hard code values, but there are some inconsistent values as well.

The two prevalent tolerances are:

One for deciding if points should be skipped.
```javascript
                       if (p1.distanceTo(prevp1) < 1e-5) {
```
One for deciding if planes are parallel.
```javascript
    CSG.Plane.EPSILON = 1e-5;
``` 
But there are tolerances that should be reviewed as well.
```javascript
        var sphereradius = bound[1] + 1e-4;
        if (absangledif < 1e-5) {
        if (Math.abs(area) < 1e-5) {
```

joostn:
I see what you mean. The rounding errors are actually a tricky problem but I've never really given it much thought. The 1e-5 originates from evanw's code, probably chosen as it is large enough to certainly smash alll rounding errors.

z3dev:
@joostn Thanks

I'll take a go at this after making a few more contributions to the test cases. I don't want to break anything in a major way.

If you would be kind enough to review the changes then I'll be happy.

z3dev:
I created PR #31 to address these tolerances. Please review if you have time. The changes are trivial but start to make sense of the tolerances.

z3dev:
The first version has been merged. So, I'm closing this.

PR #31, opened 2017-04-01
Making sense of Epsilon values

z3dev:
These changes try to make sense out of the hard coded epsilon values (tolerences) found in the library. Those are now replaced by proper static values; CSG.EPS, CSG.angleEPS, and CSG.areaEPS.

All comments, suggestions, critics are welcome. All for the better, I hope.

#29

### DIFF
--- a/csg.js
+++ b/csg.js
@@ -109,6 +109,22 @@ for solid CAD anyway.
      */
     CSG.defaultResolution3D = 12;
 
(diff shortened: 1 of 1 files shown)


kaosat-dev:
Another step in the right direction I think !
Even if we end up merging some of those , at this stage this is great !

PR #34, opened 2017-04-30
Completing Supporting Documents

z3dev:
## WORK IN PROGRESS

- [X] Update README
- [X] Add CONTRIBUTING
- [X] Document changes to API
  - Imported library from openjscad
  - Wrapped for NPM usage
  - Corrected Vector3D() in order to instantiate from anonymous objects
  - Corrected several default states for CAG.isCanonicalized
  - Corrected missing variable declarations in CAG.rectangle()
  - Added options (symmetrical) to extrudeInOrthonormalBasis and extrudeI……nPlane
  - Added CSG.cylinderElliptic()
  - Added automatic creation of API documentation via JSDOC
- [ ] Decide where to put the API documentation

Once we have this done, then I'll apply similar changes to the other libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@
 [![GitHub version](https://badge.fury.io/gh/jscad%2Fcsg.js.svg)](https://badge.fury.io/gh/jscad%2Fcsg.js)
 [![Build Status](https://travis-ci.org/jscad/csg.js.svg)](https://travis-ci.org/jscad/csg.js)
 
-> Solid modeling library (2d & 3d)
+> Solid modelling library (2d & 3d)
 
 ## Overview
 
-Constructive Solid Geometry (CSG) is a modeling technique that uses Boolean operations like union and intersection to combine 3D solids. This library implements CSG operations on meshes elegantly and concisely using BSP trees, and is meant to serve as an easily understandable implementation of the algorithm.
+Constructive Solid Geometry (CSG) is a modelling technique that uses Boolean operations like union and intersection to combine 3D solids. This library implements CSG operations on meshes elegantly and concisely using BSP trees, and is meant to serve as an easily understandable implementation of the algorithm.
 
 ## Table of Contents
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ npm run build-docs
 
 And open out/index.html using a browser.
 
+Also see the [OpenJsCad User Guide](https://en.wikibooks.org/wiki/OpenJSCAD_User_Guide).
+
 For questions about the API, please contact the [User Group](https://plus.google.com/communities/114958480887231067224)
 
 ## Contribute
--- a/README.md
+++ b/README.md
@@ -19,7 +19,6 @@ Constructive Solid Geometry (CSG) is a modeling technique that uses Boolean oper
 - [Contribute](#contribute)
 - [License](#license)
 
-
 ## Installation
 
 ```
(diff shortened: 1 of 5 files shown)


kaosat-dev:
A few thoughts:

API: 
- great idea about 'Document changes to API': I think we should put that in a changelog file :)
- side note, there are ways to autogenerate that from our git commits, but we would have to change how we merge PRS a bit
- the generated docs should be delivered with the repo, not generated manually by the user (we want to make things easy for end users) , but perhaps in a clearer manner than an 'out' folder ?

About CONTRIBUTING
- it is already "[standardized](https://github.com/blog/1184-contributing-guidelines)" on Github so we should stick to what works ie a contributing.md file
- best to avoid wiki pages, stick to md files (less of 'out of repostiory' data)

About README
- list of maintainers : yeah I don't think we need that for now ?

z3dev:
CONTRIBUTING
- Available now. Please review.
- Maybe the first section should move to the README, leaving development in CONTRIBUTING
- There are still links to WIKI pages for Reporting Issues guide, and Contributing guide.

README
- Looks pretty good.
- Shall the API generation be dropped? i think so

API DOCS
- Would be cool if GitHub could support this but...
- Maybe www.openjscad.org can house this as another set of pages?
- Rename 'out' to 'doc'?

z3dev:
Looking good.

Issues / Bugs
- Suggest adding templates then an offline document isn't required

API Documentation
- Documentation should be automatically generated for each release increment
- Where to put the documentation?. Any ideas?
  - Put the HTML/CSS on GitHub? I don't think this works
  - Put the HTML/CSS on www.openjscad.org ? Might work but we need access

kaosat-dev:
Very nice progress !
- out to doc is good
- agreed on the issue template, let's do that!
- for the docs, perhaps a good solution is using the gh-pages branch to publish it as a mini website

jsdoc.json
```diff
@@ -2,7 +2,7 @@
   "opts": { // command line options
```
kaosat-dev: Javascript comments are not valid in a json document

jsdoc.json
```diff
@@ -2,7 +2,7 @@
...
     "template": "templates/default",
     "encoding": "utf8",
-    "destination": "./out/",
+    "destination": "./doc/",
     "recurse": true,
     "readme": "README.md"
     //"tutorials": "path/to/tutorials", // same as -u path/to/tutorials
```
kaosat-dev: same here

package.json
```diff
@@ -53,7 +53,8 @@
   "devDependencies": {
     "ava": "^0.19.1",
     "babel-preset-es2015": "^6.24.1",
-    "babel-register": "^6.24.1"
+    "babel-register": "^6.24.1",
+    "jsdoc": "^3.4.3"
```
kaosat-dev: groovy :)

README.md
```diff
@@ -36,13 +35,25 @@ const csg = require('csg')
...
-TODO
+And open doc/index.html using a browser.
 
-## Contribute
+Also see the [OpenJsCad User Guide](https://en.wikibooks.org/wiki/OpenJSCAD_User_Guide).
 
 For questions about the API, please contact the [User Group](https://plus.google.com/communities/114958480887231067224)
 
-PRs accepted.
+## Contribute
```
kaosat-dev: still cannot make up my mind about where this belongs

z3dev:
API Docs

Great! Let's try to use gh-pages. It's a good solution even if we have to publish manually.

PR #36, opened 2017-05-06
Last minute changes to README

z3dev:


z3dev:
FYI, only the last commit to README.md is relevant.

PR #40, opened 2017-05-27
CAG Enhancement for toPoints()

z3dev:
So, CAG.toPoints() is not part of the library. This has been asked for by several users, and makes perfect sense. In addition, to/from and from/to points is reciprocal. Testing for this has been added.

As part of the testing, I found out that CAG.circle() was using fromSides(). Although, there's very little difference to fromPoints((), there are differences internally.
- The order of vertexes are very different
- canonicalized() is not called
So, the resulting CAG objects are really non-CAG.

Last, there were changes to the tests as the new CAG.circle() produced slightly different results.

NOTE: I dropped the new version of CSG.js into OpenJSCAD.org, rebuild, and ran all examples. Looks fine.

### DIFF
--- a/csg.js
+++ b/csg.js
@@ -6137,36 +6137,6 @@ for solid CAD anyway.
             return false;
         },
 
-        fixClosure: function() {
-            var nsides = this.sides.length;
-            if (nsides < 4) return;
-            var gapstart = 0; // v1 vertex is missing
-            for (gapstart = 0; gapstart < nsides; gapstart++) {
-                var v1 = this.sides[gapstart].vertex1;
-                var ii = 0;
-                for (ii = 0; ii < nsides; ii++) {
-                  var v0 = this.sides[ii].vertex0;
-                  if (v0.pos.x == v1.pos.x & v0.pos.y == v1.pos.y) break;
-                }
-                if (ii == nsides) break;
-            }
-            if (gapstart < nsides) {
-                var gapend = 0; // v0 vertex is missing
-                for (gapend = 0; gapend < nsides; gapend++) {
-                    var v0 = this.sides[gapend].vertex0;
-                    var ii = 0;
-                    for (ii = 0; ii < nsides; ii++) {
-                      var v1 = this.sides[ii].vertex1;
-                      if (v0.pos.x == v1.pos.x & v0.pos.y == v1.pos.y) break;
-                    }
-                    if (ii == nsides) break;
-                }
-                if (gapend < nsides) {
-                    this.sides[gapend].vertex0 = this.sides[gapstart].vertex1;
-                }
-            }
-        },
-
         expandedShell: function(radius, resolution) {
             resolution = resolution || 8;
             if (resolution < 4) resolution = 4;
--- a/csg.js
+++ b/csg.js
@@ -6002,6 +6002,24 @@ for solid CAD anyway.
             return polygons;
         },
 
+        /**
+         * Convert to a list of points.
+         * @return {points[]} list of points in 2D space
+         */
+        toPoints: function() {
+            var points = this.sides.map(function(side) {
+                var v0 = side.vertex0;
+                var v1 = side.vertex1;
+                return v0.pos;
+            });
+        // due to the logic of CAG.fromPoints()
+        // move the first point to the last
+            if (points.length > 0) {
+                points.push(points.shift());
+            }
+            return points;
+        },
+
         union: function(cag) {
             var cags;
             if (cag instanceof Array) {
--- a/csg.js
+++ b/csg.js
@@ -2228,7 +2228,7 @@ for solid CAD anyway.
         },
 
         toString: function() {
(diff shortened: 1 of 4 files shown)


csg.js
```diff
@@ -6129,14 +6129,44 @@ for solid CAD anyway.
...
-                        if (debug) { OpenJsCad.log(side0); OpenJsCad.log(side1);}
+                        if (debug) { console.log("side "+i+": "+side0); console.log("side "+ii+": "+side1);}
                         return true;
                     }
                 }
             }
             return false;
         },
 
+        fixClosure: function() {
```
kaosat-dev: Can you please explain what this does ? I do not feel wiser looking at the name and code sorry :)

csg.js
```diff
@@ -5713,21 +5719,17 @@ for solid CAD anyway.
         var center = CSG.parseOptionAs2DVector(options, "center", [0, 0]);
         var radius = CSG.parseOptionAsFloat(options, "radius", 1);
         var resolution = CSG.parseOptionAsInt(options, "resolution", CSG.defaultResolution2D);
-        var sides = [];
+        var points = [];
         var prevvertex;
-        for (var i = 0; i <= resolution; i++) {
+        for (var i = 0; i < resolution; i++) {
```
kaosat-dev: does changing to strict equality not break things ?

csg.js
```diff
@@ -6109,14 +6129,44 @@ for solid CAD anyway.
                 for (var ii = i + 1; ii < numsides; ii++) {
                     var side1 = this.sides[ii];
                     if (CAG.linesIntersect(side0.vertex0.pos, side0.vertex1.pos, side1.vertex0.pos, side1.vertex1.pos)) {
-                        if (debug) { OpenJsCad.log(side0); OpenJsCad.log(side1);}
+                        if (debug) { console.log("side "+i+": "+side0); console.log("side "+ii+": "+side1);}
```
kaosat-dev: Good change ! 
even better to use es6 template strings like 
console.log(`side ${i} ${side0}`
console.log(`side ${ii} ${side1}`

csg.js
```diff
@@ -6481,6 +6531,11 @@ for solid CAD anyway.
                     }
                     thisside = sideTagToSideMap[nextsidetag];
                 } // inner loop
+                // due to the logic of CAG.fromPoints()
+                // move the first point to the last
+                if (connectedVertexPoints.length > 0) {
```
kaosat-dev: is this to 'close' shapes ?

csg.js
```diff
@@ -6599,23 +6654,18 @@ for solid CAD anyway.
     };
 
     CAG.Side._fromFakePolygon = function(polygon) {
-        polygon.vertices.forEach(function(v) {
```
kaosat-dev: Why remove this check ?

z3dev: _fromFakePolygon is only called from fromFakeCSG, which by definition has to be in that range of Z. Why check every single vertex? The Z is controlled internally.

csg.js
z3dev: I thought we were moving backward from ES6.

csg.js
z3dev: I thought that made sense... I'll have to check again. Thanks.

csg.js
kaosat-dev: @z3dev no no no ! only for import export !! the rest of es6 is golden, works pretty much everywhere !

csg.js
kaosat-dev: ok , fine by me in that case :)

csg.js
kaosat-dev: or is this also related to improve the 'not closed' shapes issues ?

z3dev: I can take this out of this PR.

This function fixes an issue that I found with the union() function. See issue #15 
Basically, CSG.union() is broken, leaving a gap in the polygons. When converting back to CAG, that leaves a gap in the sides. This routine detects and fixes a gap.

I don't like it but...

csg.js
z3dev: Actually, it corrects the order of the sides (points).

I think this is really important a when extruding CAGs that are rotating, morphing, etc.

kaosat-dev:
Btw @z3dev  if you want to avoid headaches (copy & paste etc) in testing how things break OpenJscad.org, just use this command inside your local folder of OpenJScad.org (if csg.js is in the same 'containing' folder ie:
someFolder 
   -->csg.js folder
   -->OpenJSCAD.org folder (do this here)
)

```npm link ../csg.js/ @jscad/csg```

csg.js
kaosat-dev: ok, that is great then !
But perhaps for the future, could we also solve this by 'sorting' the points in a specific manner (using actual sort functions etc)

z3dev: Agreed. But let's save that for another discussion about what is a 100% proper CAG.

z3dev:
Done. I believe this is ready for merge now. Please review one last time.

csg.js
z3dev: It's the comparison change was because this is POINTS now, not SIDES. Can you check again?

PR #49, opened 2017-06-28
refactor(csg.js): move Path2.innerToCAG to CAGFactories. Renamed as f…

jeromew:
This a pull request regarding innerToCAG => fromPath2

### DIFF
--- a/csg.js
+++ b/csg.js
@@ -163,6 +163,7 @@ CSG.toPointCloud = require('./src/debugHelpers').toPointCloud
 const CAGMakers = require('./src/CAGFactories')
 CAG.fromObject = CAGMakers.fromObject
 CAG.fromPointsNoCheck = CAGMakers.fromPointsNoCheck
+CAG.fromPath2 = CAGMakers.fromPath2
 
 // ////////////////////////////////////
 addTransformationMethodsToPrototype(CSG.prototype)
--- a/src/CAGFactories.js
+++ b/src/CAGFactories.js
@@ -2,6 +2,7 @@ const CAG = require('./CAG')
 const Side = require('./math/Side')
 const Vector2D = require('./math/Vector2')
 const Vertex = require('./math/Vertex2')
+const Path2 = require('./math/Path2')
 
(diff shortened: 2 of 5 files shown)


z3dev:
@jeromew Big thanks for these changes.

Its great to have another set of eyes around this code. Exciting!

jeromew:
@z3dev ok I will add the comments. I must admit I lazily followed the code style in Path2.js ;-)

jeromew:
I added the documentation and squashed the commits. The doc makes it clear that getPoints returns the internal, potentially mutable points of the Path. I hesitated initially to return a copy of the points (cloning the path and returning the cloned points).

That would be slower since we add operations but I don't yet have a fine grained vision of what the state of things is in the library and it seems both of you prefer immutability.

tell me what you prefer and I will change the PR accordingly.

z3dev:
The library follows the rule of returning a NEW object, but I guess there are probably some exceptions. Is there some simple rules for return values in JavaScript? I haven't seen anything, so I assume that returns are generally "references", unless there's a specific NEW.

In this case, returning the list of points is probably best.

jeromew:
I saw many places indeed where returned objects are deep-cloned, creating a sort of immutability.
In javascript, basic types like string, booleans and integers are returned as they are. They are values and immutable in the sense that a value cannot be modified.

objects and arrays are returned as references. So modifying the elements modifies the original object. You need to deep-clone the object/array if you want something that can live its own life.

z3dev:
Thanks for the contribution,

kaosat-dev:
I know this one is merged , but thanks for the contribution :)
I am very much in favour of immutability as far as possible, but there are performance tradeoffs .
In general csg.js does seem to be on the side of returning new objects (I have been scolded by a contributor a few years back when I broke this rule :) 

This is again a subject worth more discussion, but there are ways to do rather fast object cloning with Object.assign & co these days , or the various functional programing libraries (I love Ramda), but  that works best with simpler data structures perhaps.

PR #51, opened 2017-08-13
Completed basic tests for Polygon and Polygon.Shared

z3dev:
Please review, and let me know if anything needs to change.

I didn't test solidFromSlices() as this seems to be a very special function which should be broken out from the CORE, i.e. not part of an object.


kaosat-dev:
All looks good to me ! Thanks for the work ! Great expansion of docstrings too !
Only thing I am not 100% is `checkIfConvex` but seems ok.

PR #52, opened 2017-09-10
Test cases and JSDOC comments for Path2D

z3dev:
Hope that you like these!


kaosat-dev:
Awesome PR @z3dev ! Lots of clean api documentation & updates!
My only beef is ```innerToCAG``` itself, but good catch on finding it was missing in some cases.
I think it might be possible to get rid of it/ an alternative to ```fromPoints``` I'll do some experiments.
Merging this one in the meantime , great work !

PR #54, opened 2017-09-22
Added basic tests for roundedCylinder, roundedCube, cylinderElliptic, polyhedron

z3dev:
as the title says.


kaosat-dev:
Very nice ,  it all looks good, thanks @z3dev  !

PR #63, opened 2017-11-01
fix(fixTJunctions): fixes issues with fixTJunctions

kaosat-dev:
This PR is a fix for https://github.com/jscad/scad-api/issues/26


z3dev:
FYI, the CAG goes outside the cube, which may be the reason for the failure.

kaosat-dev:
good catch @z3dev ! 
it still should not fail this way though, and its a giant mess

PR #74, opened 2017-12-20
Fix shared object without color

stuikomma:
One has to call the constructor `CSG.Polygon.Shared` either with a color or `null`. This means that you can't do `CSG.Polygon.Shared.fromObject(new CSG.Polygon.Shared(null))`.

This PR fixes that. So now you can do `new CSG.Polygon.Shared()` as well as `CSG.Polygon.Shared.fromObject(new CSG.Polygon.Shared())`.

### DIFF
--- a/src/math/Polygon3.js
+++ b/src/math/Polygon3.js
@@ -513,7 +513,7 @@ Polygon.isStrictlyConvexPoint = function (prevpoint, point, nextpoint, normal) {
  *   let shared = new CSG.Polygon.Shared([0, 0, 0, 1])
  */
 Polygon.Shared = function (color) {
-  if (color != null) {
+  if (color !== null && color !== undefined) {
     if (color.length !== 4) {
       throw new Error('Expecting 4 element array')
     }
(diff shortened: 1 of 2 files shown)


z3dev:
@stuikomma Thanks for the fixes. We will take these into the next MAJOR release, which is coming soon.

We would appreciate some test cases for these changes. I believe there is already some test cases for Polygon and Shared. See the tests folder. Can you add one or two simple tests?

kaosat-dev:
Sorry for the slow reply @stuikomma !
Good catch ! 
But instead of modifying the type of comparison operator (`!==` is prefered in all cases), the check would need to be against `undefined`

stuikomma:
Hope you enjoyed your holidays! 😃

I changed it to two strict inequality checks against `null` and `undefined` (because it would break with `null`, otherwise).

kaosat-dev:
Haha yes, thanks and you ? :)
Thanks for the tweaks ! Merging !

z3dev:
@stuikomma Welcome to JSCAD! Please feel free to contribute. There’s plenty to do and few hands.

PR #89, opened 2018-01-17
fixed typo in Vector2D documentation

ingobecker:


kaosat-dev:
Good catch & thanks for this @ingobecker ! Merging

PR #90, opened 2018-01-22
fixing angle-axis style rotation

ingobecker:
Reapplied the rotation fix.

### DIFF
--- a/src/api/ops-transformations.test.js
+++ b/src/api/ops-transformations.test.js
@@ -82,6 +82,12 @@ test('rotate (multiple items in array)', t => {
   t.deepEqual(obs.properties.sphere.center, {_x: 1.0533008150126533, _y: 1, _z: 0.9436934847150743})
 })
 
+test('rotate (single item angle axis style)', t => {
+  const op1 = cube()
+  const obs = rotate(Math.PI, [0, 1, 0], op1)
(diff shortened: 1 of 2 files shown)


src/api/ops-transformations.test.js
```diff
@@ -225,7 +231,7 @@ test('expand (single item)', t => {
   const op1 = cube()
   const obs = expand(10, 5, op1)
 
-  t.deepEqual(obs.polygons[0].vertices[0], {pos: {_x: -10, _y: 0, _z: 0}, tag: 25909})
```
ingobecker: This test failed caused by a differing tag value. I don't excatly know what a tag is used for but it looks like some kind of uniqe value to me. Hope changing it doesn't have side effects and is ok.

kaosat-dev: That is an unfortunate side effect for the internals of CSG objects (there is an ever increasing `tag` counter that is used by some of the boolean operation algorithm to distinguish or fuse together polygons. It is very clunky and quite annoying for tests since the tag value can vary depending on what tests ran before. 
So no worries :)

kaosat-dev:
Thanks a lot for opening this again here @ingobecker  :) much appreciated! 
I will take a closer look & merge if all is good !

kaosat-dev:
All good, merging ! Thanks @ingobecker (and sorry for the delay!)

PR #91, opened 2018-01-28
fix(CAG): added back missing CAG.fromCompactBinary

kaosat-dev:
This PR fixes a MAJOR regression introduced in the last release that broke CAG.fromCompactBinary


z3dev:
Cool.
So, were the test cases correct?

kaosat-dev:
@z3dev  they were disabled as the object 'tags' were causing issues, but I added a simpler helper to sort that side out, and they worked fine :)

PR #94, opened 2018-02-07
added test for complex edge-case subtract with two example solids

plnralex:
This test is set to failing but will not fail when const EPS is set to 0.0001. Testing for edge cases on csg.subtract() can be done by comparing the bounding distance before and after the subtraction. After subtraction the result should not be bigger then before. Equal or smaller would be acceptable results.


kaosat-dev:
Thanks a lot for this @plnralex !

PR #97, opened 2018-03-03
Correcting center() functionality

z3dev:
WORK IN PROGRESS DO NOT MERGE !!

These changes correct the various flaws of the current center functionality.

- [x] Correct centering of objects
- [x] Update core objects to alias API
- [x] Add or update test suites
- [x] Add or update JSDOC
- [x] Update user guide (new wiki only)

### DIFF
--- a/src/api/center.js
+++ b/src/api/center.js
@@ -1,18 +1,21 @@
 const toArray = require('../core/utils/toArray')
 
 /**
- * centers the given object(s) using the given options (if any)
+ * Centers the given object(s) using the given options (if any)
  * @param {Object} [options] - options for centering
  * @param {Array} [options.axes=[true,true,true]] - axis of which to center, true or false
  * @param {Array} [options.center=[0,0,0]] - point of which to center the object upon
  * @param {Object|Array} objects - the shape(s) to center
+ * @return {Object|Array} objects
  *
- * TODO : Add addition 'methods' of centering; midpoint, centeriod
+ * @example
+ * let csg = center({axes: [true,false,false]}, sphere()) // center about the X axis
  */
 const centerV2 = function (options, objects) {
   const defaults = {
     axes: [true, true, true],
     center: [0, 0, 0]
+  // TODO : Add addition 'methods' of centering; midpoint, centeriod
   }
   options = Object.assign({}, defaults, options)
   const {axes,center} = options
--- a/src/api/center.js
+++ b/src/api/center.js
@@ -1,11 +1,11 @@
 const toArray = require('../core/utils/toArray')
 
-/** NOTE: this is not functional YET !!
- * centers the given object(s) on the given axis
+/**
+ * centers the given object(s) on the given axes
  * @param {Object|Array} object(s) the shapes to center
  * @param {Object} options
  */
-const center = (objects, options) => {
+const center = function (objects, options) {
   const defaults = {
     axes: [1, 1, 1]
   }
@@ -15,9 +15,11 @@ const center = (objects, options) => {
 
   const results = objects.map(function (object) {
     let b = object.getBounds()
-    return object.translate(axes.map(function (a) {
-      return cAxes.indexOf(a) > -1 ? -(b[0][a] + b[1][a]) / 2 : 0
-    }))
+    let offset = [0,0,0]
+    if (axes[0]) offset[0] = 0 - (b[0]._x + ((b[1]._x - b[0]._x) / 2))
+    if (axes[1]) offset[1] = 0 - (b[0]._y + ((b[1]._y - b[0]._y) / 2))
+    if (axes[2]) offset[2] = 0 - (b[0]._z + ((b[1]._y - b[0]._y) / 2))
+    return object.translate(offset)
   })
   // if there is more than one result, return them all , otherwise a single one
   return results.length === 1 ? results[0] : results
--- a/src/api/center.js
+++ b/src/api/center.js
@@ -5,7 +5,7 @@ const toArray = require('../core/utils/toArray')
  * @param {Object|Array} object(s) the shapes to center
  * @param {Object} options
  */
-const center = function (objects, options) {
+const centerV2 = function (objects, options) {
   const defaults = {
     axes: [1, 1, 1]
   }
@@ -25,4 +25,4 @@ const center = function (objects, options) {
   return results.length === 1 ? results[0] : results
 }
 
-module.exports = center
+module.exports = centerV2
--- a/src/api/center.js
+++ b/src/api/center.js
@@ -11,7 +11,7 @@ const toArray = require('../core/utils/toArray')
  * @example
  * let csg = center({axes: [true,false,false]}, sphere()) // center about the X axis
  */
-const centerV2 = function (options, objects) {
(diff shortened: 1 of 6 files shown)


z3dev:
@kaosat-dev Please review.

Things that bother me...
- Should the API center.js code in ops-transforms.js? Why have another piece of code?
- The option (axis to center) has changed. Is this moving towards V2? or way?

src/api/center.js
```diff
@@ -15,12 +15,14 @@ const center = (objects, options) => {
 
   const results = objects.map(function (object) {
     let b = object.getBounds()
```
kaosat-dev: a more explicit/clear variable name than 'b'  would be better :)

src/api/center.js
```diff
@@ -1,11 +1,11 @@
...
 
-/** NOTE: this is not functional YET !!
- * centers the given object(s) on the given axis
+/**
+ * centers the given object(s) on the given axes
  * @param {Object|Array} object(s) the shapes to center
  * @param {Object} options
  */
-const center = (objects, options) => {
+const centerV2 = function (objects, options) {
```
kaosat-dev: why rename it ?

src/api/center.js
```diff
@@ -15,12 +15,14 @@ const center = (objects, options) => {
 
   const results = objects.map(function (object) {
     let b = object.getBounds()
-    return object.translate(axes.map(function (a) {
-      return cAxes.indexOf(a) > -1 ? -(b[0][a] + b[1][a]) / 2 : 0
-    }))
+    let offset = [0,0,0]
+    if (axes[0]) offset[0] = 0 - (b[0]._x + ((b[1]._x - b[0]._x) / 2))
```
kaosat-dev: seems good to me !
but why use the _x & _y etc instead of just x & y ? (is it the whole accessors issue?)

src/api/ops-transformations.js
```diff
@@ -135,6 +135,9 @@ function center (axis, ...objects) { // v, obj or array
       object = object.union(_objects[i])
     }
   }
+  if (axis.constructor !== Array) {
```
kaosat-dev: from what I have seen , apparently the best way to test for an Array is `Array.isArray()`

src/api/ops-transformations.test.js
```diff
@@ -185,7 +185,7 @@ test('transform (multiple items, 2d , translation)', t => {
 test('center (single item)', t => {
   const op1 = cube()
   const obs = center(true, op1)
-  t.deepEqual(obs.properties.cube.center, {_x: 0.5, _y: 0.5, _z: 0.5})
+  t.deepEqual(obs.properties.cube.center, {_x: 0, _y: 0, _z: 0})
```
kaosat-dev: yay :) a logical result !!

src/core/CAG.js
```diff
@@ -101,6 +102,12 @@ CAG.prototype = {
     return fromSides(newsides)
   },
 
+  // ALIAS !
+  center: function (axes) {
+    axes = [axes[0],axes[1],false] // just be sure
```
kaosat-dev: I do not quite understand this one ?  was it what the old version was doing ? if yes, wow , weird

z3dev: Hmm... good point. I’ll just pass the given argument through.

src/api/center.js
z3dev: I didn’t like the fact that center() was calling center().

src/api/ops-transformations.js
z3dev: Yup. I did the research too.
We should be consistent going forward.

src/api/center.js
z3dev: centerV2 = function (options, objects) is correct

kaosat-dev: but centerV2 is just awkward :)
one center is a method attached to prototype, the other one a simple function, I see no issues with it to be honest :)

z3dev: True. CSG.center() calls centerV2()

So does API center(). See ops-transformations.js

z3dev:
Updates completed.

An addition option was added called 'center' :), which allows the object(s) to be centered at any given point. COOL

z3dev:
Another thought... how about adding these? Nice simple functions. The rotateX() rotateY() rotateZ() exist today.

centerX()
centerY()
centerZ()

kaosat-dev:
Ugh, I really dislike those to be honest : I add YET MORE API , and they are only marginally shorter than 
center([true, false, false]) : I would honestly rather we reduce the API , not expand it

z3dev:
Sorry one more question before wrapping this PR for release.

centerV2() accepts:
- options: axes=[true,true,true]
- objects: object or array of object

Should this be accepting or or more objects using the REST paramater passiing? ...objects?

src/api/center.js
```diff
@@ -1,18 +1,21 @@
...
  *
- * TODO : Add addition 'methods' of centering; midpoint, centeriod
+ * @example
+ * let csg = center({axes: [true,false,false]}, sphere()) // center about the X axis
  */
 const centerV2 = function (options, objects) {
   const defaults = {
     axes: [true, true, true],
     center: [0, 0, 0]
+  // TODO : Add addition 'methods' of centering; midpoint, centeriod
```
kaosat-dev: Great idea !

kaosat-dev:
hmm  I  think it should accept both a single object or an array, still not sure about  center(options, obj1, obj2) , then again it would be coherent with the rest of the api ?

src/api/center.js
z3dev: i have to finish those 'measurements' first. those doen't exist today. add another thing to the list.

z3dev:
OK. If there are no more changes to the parameters passing then this PR is finished. Any last review or comments?

kaosat-dev:
yes a few actually : 
- I still dislike calling it centerV2 :)
- I am confused about one main thing : this version of center (after/ and even before your fixes) is now used instead of the nasty mutators : GOOD ! 
   * but it is not exposed as part of the API : (because we still have the center() in ops-transformations)
   * we should only have one function , but I checked and the current parameter ordering for API center() is bad , ie inverted conpared to the new sexy center() , so the question is:
     * do we merge this 'as -is', and deal with having a single function later but not break the API (so no version number issues)
    * do we merge to a 'V2 branch' as we talked about, thus we would be able to have a single function, much cleaner, and a better API , much better

z3dev:
That’s the heart of the issue. How are we going to implement the transition to V2?

Whoever we switch over, it should probably be one BIG BANG.
- removal of the aliases on the CSG/CAG/etc objects
- Decommission the old SCAD like API
- Implement V2 API

The appeal of a V2 branch is good. We can work away from the current, and work in concistency and pureity.

Also, we shouldn’t be scared to make a few small increments to V1 API if that helps to make the BIG BANG easier. For example, removing the mutators.

I’ll rename centerV2 but like you, I don’t like two versions of center() in the API. Maybe we just have to live with this for a while.

z3dev:
Thanks. I think this small API change is warranted.

Just let me change centerV2() to center()

z3dev:
Last change completed. Updated new WIKI as well.

PR #102, opened 2018-04-22
remove quoted paths for windows

lautr3k:
The quotes must be escaped under windows, as there are no spaces in the paths, they are not essential.


kaosat-dev:
ok thanks, tests are passing, merging right away

PR #103, opened 2018-04-23
Fix/from nested points array

lautr3k:
Sorry, I did not test well, here are some fixes/improvements.


kaosat-dev:
Super duper thanks ! It looks good to me !

PR #106, opened 2018-04-29
Feature/text()

lautr3k:
jscad

```javascript
var roboto = include('roboto-2014/Roboto-Regular.ttf');
var stylus = include('Deftone-Stylus/Deftone-Stylus.ttf');

function main(params) {
  var f1 = font(roboto);
  var f2 = font(stylus);

  var t1 = text({ text: 'OpenJSCAD', font: f1 });
  var t2 = text({ text: 'OpenJSCAD', font: f1, center: [true, true], letterSpacing: 10 });
  var t3 = text({ text: 'OpenJSCAD', font: f2, center: [true, true], width: 100 });
  var t4 = text({ text: 'OpenJSCAD', font: f2, center: [true, true], width: 100, height: 40 });

  return [
    t1.translate([0,  50, 0]),
    t2.translate([0,  30, 0]),
    t3.translate([0,   0, 0]),
    t4.translate([0, -50, 0])
  ];
}
```


kaosat-dev:
This rocks :) thanks ! I'll take a look

kaosat-dev:
@lautr3k : overall super good :) ! 
As discussed yesterday Only change that would be important is having the actual text/string as a seperate parameter not part of the options ie  
text(options, textString)

```const t1 = text({font: f1 }, 'OpenJSCAD')```

Something else I thought of today while taking another look at your PR: 
- perhaps width/height should be collapsed into a single 'size' parameter (2d array ?)

z3dev:
What are the default options? The routine should return something meaningful for simple cases.
- default font? Size?
- return empty CAG?

lautr3k:
Default font size is 10 if not set in `options.height`
For the default font, I will use vector_text() as fallback

z3dev:
That sounds reasonable.
Just wondering, is there a better “vector” font available? Could another default font be created? Or found? The current vector font is very limited.

lautr3k:
@z3dev limited in terms of what? number of characters supported? encoding?

lautr3k:
web -> drag/drop fonts files then `var myFont = include('myFont.ttf')` 
desktop -> local fonts can be used or drag/drop

z3dev:
Sounds really easy. Also, there's a huge list of fonts already available.

So... Can does HTML / CSS standard specify a standard font? One that every OS / Browser must have?

lautr3k:
No, opentype need a font file to work.

z3dev:
@lautr3k General question... we need a few more eyes for reviews and opinions for design questions... can we ask (you) for reviews / comments when relevant?

lautr3k:
Of course, feel free to ask. But note that I have already started another refactoring of the textes functions in another branch. I rushed on that one and mixed too much stuff at once. I would like to do it again in two stages (PR):
1 - Refactoring of `vector_char()` & `vector_text()` : options first, font-size, line-height, letter-spacing, optimization.
2 - Add the `text()` function

z3dev:
@lautr3k Any issues with the new PR? I'm looking forward to playing with the new functionalty.

PR #108, opened 2018-06-17
Fix and Tests for CAG measurements and Poly3 measurements

z3dev:
Additional tests were added to verify and test area calculations for both CAG and Polygon3.

Issues with Polygon3 were found when the polygon was 'concave' in shape. Fixes were made to use a faster routine for area calculations.

Note: measureArea.js was added for the new routine, which should be easy to integrate into V2.

### All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [x] Does your submission pass tests? (Including new tests)


z3dev:
These are now part of V2 so closing.

PR #110, opened 2018-07-07
V2 fix for vec3.fromVarious()

z3dev:
This function is required for many of the tests to suceed. The fixes include:
- reworked to use arguments
- reworked to check return values and return proper values
- reworked object conversions, and removed non-V2 objects

I also added IsFloat() to util.js as math classes will need this.

Let me know how this looks.

### DIFF
--- a/src/core/math/utils.js
+++ b/src/core/math/utils.js
@@ -1,4 +1,8 @@
 const radToDeg = radians => radians * 114.59155903
 const degToRad = degrees => degrees * 0.00872665
 
-module.exports = {radToDeg, degToRad}
+const IsFloat = function (n) {
+  return ! (isNaN(n) || (n === Infinity) || (n === -Infinity))
+}
+
+module.exports = {radToDeg, degToRad, IsFloat}
--- a/src/core/math/utils.js
+++ b/src/core/math/utils.js
@@ -1,8 +1,4 @@
 const radToDeg = radians => radians * 114.59155903
 const degToRad = degrees => degrees * 0.00872665
 
-const IsFloat = function (n) {
-  return ! (isNaN(n) || (n === Infinity) || (n === -Infinity))
-}
-
-module.exports = {radToDeg, degToRad, IsFloat}
+module.exports = {radToDeg, degToRad}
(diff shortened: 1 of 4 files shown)


z3dev:
@kaosat-dev please review, and if happy then please merge into V2

src/core/math/utils.js
```diff
@@ -1,4 +1,8 @@
 const radToDeg = radians => radians * 114.59155903
 const degToRad = degrees => degrees * 0.00872665
 
-module.exports = {radToDeg, degToRad}
\ No newline at end of file
+const IsFloat = function (n) {
```
kaosat-dev: isFloat is already defined src/core/utils/typeChecks.js

src/core/math/vec3/fromVarious.js
```diff
@@ -14,68 +12,68 @@ const {IsFloat} = require('../utils')
...
 function fromVarious (...params) {
   let out = new Float32Array(3)
-  if (params.length === 3) {
-    out[0] = parseFloat(x)
-    out[1] = parseFloat(y)
-    out[2] = parseFloat(z)
-  } else if (params.length === 2) {
-    out[0] = parseFloat(x)
-    out[1] = parseFloat(y)
+  if (arguments.length === 3) {
```
kaosat-dev: please use params instead of arguments, not the same :)

src/core/math/vec3/unit.js
```diff
@@ -0,0 +1,14 @@
...
+const fromValues = require('./fromValues')
+const length = require('./length')
+
+/**
+ * Calculates the unit length of a vec3
+ *
+ * @param {vec3} a vector to calculate unit length of
+ * @returns {Number} unit length of a
+ */
+function unit (a) {
```
kaosat-dev: - I feel stupid, but why do we need to pass in a length to a unit vector ?

kaosat-dev: another remark but that is 100% my fault, the very short variable names originating from gl-vecXX are too short and non comprehensible for my taste (I have started to replace some). so feel free to replace `a`  & `l` with something more clear :)

z3dev: Actually, i'm not 100% as to which 'length' is being observed, i.e. what is 'w'?. I think this is the "unit vector" which represents the normal and tagent to perpendicular, but I could be wrong.

http://mathworld.wolfram.com/NormalVector.html

Do you know?

PR #112, opened 2018-07-29
renovation() : V2 version of plane, see Plane.js for notes

z3dev:
This isn't perfect or tested, but implements V2 API, compiles, and works with some other test cases.

I think this also completes the math functions / data structures.


kaosat-dev:
@z3dev sorry for the late reply, I am more conflicted about this one : plane is essentially vec4 with a few extra helper functions, I think we should reuse most of those (we can keep a seperate plane folder with the extra functions/modules, and just re-use/ and re-export the ones that are identical to vec4, without code duplication): we can either merge & change , or change before the merge, what do you think ?

z3dev:
Thanks. I’ll revisit and reuse vec4.
For vec4, I will extend with a few more functions, and make it similar to vec3.
For plane, I will only retain the special functions. And update the index.js to reuse vec4 functions.

kaosat-dev:
ok @z3dev , thanks do you want to merge this first or do you want to change the PR ?

z3dev:
I will update this pull request.

z3dev:
@kaosat-dev please review.

i think this is ready for merging into V2.

PR #113, opened 2018-08-12
Math Test Suite

z3dev:
WORK IN PROGRESS

Adding test suites for math functionality
 - [ ] mat4
 - [ ] plane
 - [ ] utils.js
 - [ ] vec2
 - [ ] vec3 IN PROGRESS
 - [ ] vec4


z3dev:
I believe that most of the functionality is ready for testing.

I have created specific test cases for each function found in the object directories. Let me know how this looks, as I think this will make smaintenance and further testing very easy.

z3dev:
@kaosat-dev please review

the test suite for vec3 is complete, with coverage of 100% for everything but fromVarious()

i want to remove fromVarious(). this is a core math class, and not meant for common use by users. if necessary then specialized functions can be added when necessary. in addition, the API and the GEOMETERY functions can convert 'various' representations to vec3 functions. we should keep vec3 clean.

z3dev:
the test suite for vec4 is complete, with coverage of 100% for everything.

src/core/math/vec3/index.js
```diff
@@ -19,10 +19,10 @@ module.exports = {
   min: require('./min'),
   multiply: require('./multiply'),
   negate: require('./negate'),
-  normal: require('./normal'),
+//  normal: require('./normal'),
   normalize: require('./normalize'),
-  randomNonParallel: require('./randomNonParallel'),
-  rotate: require('./rotate'),
+  random: require('./random'),
```
kaosat-dev: looking at the code for 'randomNonParallel' it seems not random at all , we need a better name

src/core/math/vec3/lerp.js
```diff
@@ -5,7 +5,7 @@ const create = require('./create')
  * Performs a linear interpolation between two vec3's
  *
  * @param {vec3} out the receiving vector
- * @param {Number} t interpolation amount between the two inputs
+ * @param {Number} t interpolant (0.0 to 1.0) applied between the two inputs
```
kaosat-dev: are you sure about this ? I do not get the word 'interpolant' , interpolation amount seems clearer
but good addition of the 0.0 to 1.0

src/core/math/vec3/lerp.js
```diff
@@ -15,7 +15,7 @@ function lerp (...params) {
   let t
   let a
   let b
-  if (params.length === 2) {
+  if (params.length === 3) {
```
kaosat-dev: nice catch

src/core/math/vec3/min.test.js
```diff
@@ -0,0 +1,48 @@
+const test = require('ava')
+const {min, fromValues} = require('./index')
+
+test('vec3: min() called with two parameters should return a vec3 with correct values', t => {
```
kaosat-dev: was min not removed from the index ? or was that just commented out for a while ?

src/core/math/vec3/random.js
```diff
@@ -1,9 +1,9 @@
-module.exports = randomNonParallel
+module.exports = random
 
 const normalize = require('./normalize')
 const fromValues = require('./fromValues')
 
-function randomNonParallel (input) {
+function random (input) {
```
kaosat-dev: same remarks as in the index
I think it should be called 'createNonParallel' or something ?

kaosat-dev:
please leave fromVarious in place for now, it is needed in many places, and I believe it would be premature to remove it at this stage.

src/core/math/vec4/from.test.js
```diff
@@ -0,0 +1,16 @@
+const test = require('ava')
+const {fromScalar, fromValues, toString} = require('./index')
+
+test('vec4: fromScalar() should return a new vec4 with correct values', t => {
```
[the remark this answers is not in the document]
z3dev: Not sure, but the test completes fine now. :)

src/core/math/vec3/index.js
z3dev: I’ll review the old code. I believe this is used by plane, and needs to be random, i.e. making a new vector randomly.

src/core/math/vec4/create.test.js
```diff
@@ -1,8 +1,8 @@
 const test = require('ava')
-const {create} = require('./index')
+const {create, fromValues} = require('./index')
 
 test('vec4: create() should return a vec4 with initial values', t => {
   const obs = create()
-  const exp = Float32Array.from([0,0,0,0])
+  const exp = fromValues(0, 0, 0, 0)
```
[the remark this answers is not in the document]
z3dev: This is the base test for TYPE. Everything else is returned b6 this function is based on what this create() function returns. There are now two (2) very TYPE specific functions.
- create()
- fromValues()

Everything else should be using those functions.

src/core/math/vec3/abs.test.js
```diff
@@ -0,0 +1,16 @@
+const test = require('ava')
+const {abs, fromValues} = require('./index')
+
+test('vec3: abs() should return a vec3 with positive values', t => {
+  const obs1 = abs([0, 0, 0])
+  const exp1 = fromValues(0, 0, 0)
```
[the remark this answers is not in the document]
z3dev: The fromValues() is a base function... it must work... period

src/core/math/vec3/lerp.js
z3dev: I’m not a mathematician, but that’s the name given by the website. 😃

src/core/math/vec3/random.js
z3dev: Well... dot and cross return a new vector, but those are not named...

createCross()
createDot()

Im flexible but we should be consistent.

src/core/math/vec3/abs.test.js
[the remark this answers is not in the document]
kaosat-dev: of course it must work, but UNIT tests this low level by definition should not rely on other parts of the API  to test the validity

src/core/math/vec3/random.js
kaosat-dev: ok, so err 'nonParallel' ?

src/core/math/vec3/abs.test.js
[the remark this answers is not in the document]
z3dev: Which would mean replacing all fromValues() with Float32Array.from([0,0,0,0])

I don’t think that would be a good choice.

z3dev:
@kaosat-dev please review the new helper function; compareVectors(). take note that it does not compare TYPES, only VALUES by operator ==.

For an example, see vec3/unit.test.js, which now uses compareVectors(). The test cases are easier to ready, and comparisons do not relie on the internals of vec3. However, I'm a little concerned about the use of high precions numbers for the comparison. We may hit some cases where we need to compare nearlyEqual  values. TBD

compareVectors(vec1, vec2, epsilon)

kaosat-dev:
looking very good so far, but yeah I am also pretty sure we will have to use nearlyEqual a lot too :) so your suggestion makes sense !

src/core/math/vec3/unit.test.js
```diff
@@ -1,56 +1,48 @@
 const test = require('ava')
 const {unit, fromValues, toString} = require('./index')
 
+const {compareVectors} = require('../../../../test/helpers/index')
```
kaosat-dev: very cool!
btw you can leave out the index btw, node looks for 'index.js' files in folders directly, so you can just use const {compareVectors} = require('../../../../test/helpers')

src/core/math/vec3/normalize.test.js
```diff
@@ -1,46 +1,40 @@
 const test = require('ava')
 const {normalize, fromValues} = require('./index')
 
+const {compareVectors} = require('../../../../test/helpers/index')
```
kaosat-dev: all looks good to me!!

PR #116, opened 2018-08-27
Add a method to determine the general turn direction of a path.

pentacular:
This will help in converting compound svg paths containing holes.

https://github.com/jscad/io/issues/71

### DIFF
--- a/src/core/math/Path2.js
+++ b/src/core/math/Path2.js
@@ -162,15 +162,17 @@ Path2D.prototype = {
   },
 
   /**
-   * Determine the overall turn of a path.
-   * @returns {String} One of 'clockwise', 'counter-clockwise', or 'straight'.
+   * Determine the area of a path were it to describe a closed polygon.
+   * see: http://mathworld.wolfram.com/PolygonArea.html
+   * @returns {Number}
    */
   getArea: function () {
-    let points = this.points;
-    let area = 0
-    for (let last = 0, current = 1; current < this.points.length; last++, current++) {
-      area += (points[last].x - points[current].x) * (points[last].y - points[current].y);
-    });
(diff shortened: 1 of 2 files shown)


z3dev:
Just a few thoughts... the function is calculating the area of a closed path, so what not call it that getArea(). The IO library can do the simple checks for < > = 0

pentacular:
Good idea.

pentacular:
Oops -- I hit the wrong button.

I changed it to expose getArea() and found where the tests live. :)
Please take a look.

kaosat-dev:
the tests are very desciptive, well done @pentacular  👍 
about the feature itself: 
- very useful to determin cw vs ccw 👍 
- but to be honest it feels a bit hackish to have negative areas 'mean' something, because there are no such things as negative areas in maths, so perhaps some other name would be better @z3dev :)
I honestly liked the previous name much better: it was very clear what it did :)

kaosat-dev:
no worries @z3dev . I had to look it up just to be sure, but given our focus on clear, logical things for V2 we might want to avoid anything iffy ;)

pentacular:
Ok, that should be up to date as getTurn -- let me know what you think.

kaosat-dev:
@pentacular we are going to merge this one, thanks for the work ! :)
but a heads up : from today onwards we will not be accepting any PRs for the V1 version, as it means a lot more work for us to transfer the V1 changes to the very different V3

pentacular:
Thanks.

Do you mean the V2 branch?
I'll see if it works with my codebase.

kaosat-dev:
It won't haha the v2 branch is in very early stages. But we will incorporate your changes there don't worry :-)

Issue #17, opened 2017-02-17
Issue with core Matrix4x4.translation function


z3dev:
_From @rasmuskr on April 27, 2016 22:9_

Hey

I am pretty sure that the function CSG.Matrix4x4.translation is broken from:
https://github.com/Spiritdude/OpenJSCAD.org/blob/master/csg.js#L3586

```
// Create an affine matrix for translation:
    CSG.Matrix4x4.translation = function(v) {
        // parse as CSG.Vector3D, so we can pass an array or a CSG.Vector3D
        var vec = new CSG.Vector3D(v);
        var els = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, vec.x, vec.y, vec.z, 1];
        return new CSG.Matrix4x4(els);
    };
```

The matrix should look like this

```
var els = [1, 0, 0, vec.x, 0, 1, 0, vec.y, 0, 0, 1, vec.z, 0, 0, 0, 1];
```

I may fix this and open a PR but dont have the tools at the computer I am currently on.

_Copied from original issue: Spiritdude/OpenJSCAD.org#148_

z3dev:
@rasmuskr Can you open an issue against CSG.js at the following repository? That's the original repository and you'll get faster answers there.

https://github.com/joostn/OpenJsCad

kaosat-dev:
Should it actually not be dealt with here @z3dev  (so the your last comment change to reflect this ?)

z3dev:
The tool brought my original comment with this issue. I believe that this is the new home of CSG.js.

Would this be better under another repository?

kaosat-dev:
No not that is ok , this makes sense in this repo, I just was not sure if the comment was part of the 'migrated' texts :)

z3dev:
Understood. Definitely hard to tell if the comment was original or transferred via the tool. Going forward, i'll add a small comment to the bottom of transferred issues.

z3dev:
V1 Matrix4x4.translation() and V2 mat4.fromTranslation() are the same, and correct.

The original request may have confused a matrix for translation VERSUS the matrix math translate by a vertex.

I'm closing this.

PR #119, opened 2018-09-08
Math Test Suite Part 2

z3dev:
WORK IN PROGRESS

PART 2 of the test suites for math functionality
 - [ ] mat4 IN PROGRESS, 52.61%, ISSUES see comments
 - [X] plane DONE, 86.36 % (no tests for fromData/fromObject/fromPoints, otherwise 100%)
 - [X] vec2 DONE, 90.84 % (no tests for fromVarious, otherwise 100%)
 - [X] vec3 DONE, 89.90 % (no tests for fromVarious), otherwise 100%
 - [X] vec4 DONE, 100.00 % coverage

Need to add tests for issue #17


z3dev:
@kaosat-dev please review quickly as there were fixes for vec2. please take a look at each of the tests for the associated fixes. cheers

src/core/math/vec2/cross.js
```diff
@@ -24,7 +24,7 @@ function cross (...params) {
   }
   let z = a[0] * b[1] - a[1] * b[0]
   out[0] = out[1] = 0
-  out[2] = z
+  out[1] = z
```
kaosat-dev: err no a cross product only exists in 3d space, and you want to set the third component (at index 2) to Z , not at index 1

src/core/math/vec2/dot.test.js
```diff
@@ -0,0 +1,24 @@
+const test = require('ava')
+const {dot, fromValues} = require('./index')
+
+test('vec2: dot() should return correct values', t => {
```
kaosat-dev: nice and simple !

src/core/math/vec2/distance.test.js
```diff
@@ -0,0 +1,30 @@
+const test = require('ava')
+const {distance, fromValues} = require('./index')
+
+const {nearlyEqual} = require('../../../../test/helpers/index')
+const {EPS} = require('../../constants')
```
kaosat-dev: are you sure about the use of EPS here ? we have not looked at that one in  along while

src/core/math/vec2/cross.test.js
```diff
@@ -0,0 +1,81 @@
+const test = require('ava')
```
kaosat-dev: nice all !

src/core/math/mat4/mirror.js
```diff
@@ -7,7 +7,7 @@ const fromValues = require('./fromValues')
  * create an affine matrix for mirroring into an arbitrary plane:
```
kaosat-dev: mirror is complicated, our old & new implementations differ quite a bit

src/core/math/vec2/cross.js
z3dev: Understood, but a vec2 function returning a vec3 result just seems wrong. How about removing this from vec2?

src/core/math/vec2/distance.test.js
z3dev: I’ve been using EPS for all the nearlyEqual tests, thinking that this would be a better value then just something arbitrary. We can review EPS when working on the geometry / BSP trees, which is where EPS becomes important.

src/core/math/mat4/mirror.js
z3dev: I’ve been testing for consistent results as well. If we need to revisit mirror then now is the time. The math functions are the heart and soul of the library.

Do you have an example? I’m sure the API is broken, so we can skip that level.

src/core/math/vec2/cross.js
kaosat-dev: no, that is totally normal, that is the maths for cross product, not something we want to change :) there are plenty of functions returning scaler (float) values aswell, it is all fine

src/core/math/mat4/mirror.js
kaosat-dev: not off the top of my head but it is essential that we keep both implementations in the file for now, until they can be unified/ merged down the line

z3dev:
@kaosat-dev please take another look, as I corrected vec2.cross. And added mat4 tests.

THIS NEEDS ATTENTION.. There are some unnecessary functions in mat4.
leftMultiplyVec2.js
leftMultiplyVec3.js
rightMultiplyVec2.js
rightMultiplyVec3.js

mat4.rightMultiply...
V1 uses only helpers.rightMultiply1x3VectorToArray(). If rightMultiply is required then we can add to vec3. So, I think the mat4 functions are never used. Can I drop these in mat4?

mat4.leftMultiply...
V1 Vector2 and Vector3 uses to transform vectors using Matrix4. Why these classes have two functions with the same functionality... I have no idea.

Also, V2 vec2.transformMat4 and vec3.transformMat4 seem to be the new V2 functions. But I think we should review closely. There seem to be slight differences in return values.

z3dev:
@kaosat-dev Please review again. I think this PR should be merged as there were a number of changes to bring the functions up to ES6, and JSCAD standards for code.

I will continue on via another PR, as there will be a number of changes in order to make functions consistent across the objects, i.e. further changes to use (...params)

src/core/math/mat4/clone.js
```diff
@@ -1,28 +1,18 @@
...
-  out[9] = a[9]
-  out[10] = a[10]
-  out[11] = a[11]
-  out[12] = a[12]
-  out[13] = a[13]
-  out[14] = a[14]
-  out[15] = a[15]
-  return out
+const clone = (a) => {
+  return fromValues(
```
kaosat-dev: smaart !

src/core/math/mat4/create.js
```diff
@@ -1,12 +1,10 @@
...
-
 /**
  * Creates a new identity mat4
  *
  * @returns {mat4} a new 4x4 matrix
  */
-function create () {
-  let out = new Float32Array(16)
+const create = () => {
+  const out = new Float32Array(16)
```
kaosat-dev: good use of const !

src/core/math/mat4/equals.js
```diff
@@ -5,7 +5,7 @@
  * @param {mat4} b The second matrix.
  * @returns {Boolean} True if the matrices are equal, false otherwise.
  */
-function exactEquals (a, b) {
+const exactEquals = (a, b) => {
```
kaosat-dev: yes !

src/core/math/mat4/equals.test.js
```diff
@@ -0,0 +1,56 @@
+const test = require('ava')
+const { equals, fromValues } = require('./index')
+
+test('vec3: equals() should return correct booleans', (t) => {
```
kaosat-dev: correct booleans ?

src/core/math/mat4/fromRotation.js
```diff
@@ -38,9 +38,9 @@ function fromRotation (...params) {
   y *= len
   z *= len
 
-  s = Math.sin(rad)
-  c = Math.cos(rad)
-  t = 1 - c
+  const s = Math.sin(rad)
```
kaosat-dev: good change

src/core/math/plane/splitLineBetweenPoints.js
```diff
@@ -1,17 +1,18 @@
...
   let labda = (plane[3] - vec3.dot(plane, p1)) / vec3.dot(plane, direction)
-  if (isNaN(labda)) labda = 0
+  if (Number.isNaN(labda)) labda = 0
   if (labda > 1) labda = 1
   if (labda < 0) labda = 0
   const result = vec3.plus(p1, vec3.times(direction, labda))
   return result
 }
+
+module.exports = splitLineBetweenPoints
```
kaosat-dev: good change (in general , regarding all these, good idea to move them after the declaration)

PR #122, opened 2018-10-15
refactor & upgrade (V2)

kaosat-dev:
This PR updates & overhauls a lot of the 2d (and to a lesser extent 3d) shapes, geometries etc for V2
It is just one step along the way.


kaosat-dev:
@z3dev this is ready for merge: a lot of things are still work in progress , but a lot of things have been worked on & improved.
I also tweaked a few things for Mat4 , & made sure only the new tests are run, so we can actually get some 'green' builds :)

z3dev:
👍🏻

PR #126, opened 2018-10-28
Rewrites of functions to remove reliance on tags

z3dev:
WORK IN PROGRESS

### SHAPE2
outlineToPaths
- [X] Convert to V2 math object
- [X] Rewrite to remove getTag
- [ ] Write test suite IN PROGRESS

toCompactBinary (Changes to be completed with rewrite of fromCompactBinary)

### SHAPE3
expand
- [ ] Convert to V2 math objects
- [ ] Rewrite to remove getTag
- [ ] Write test suite

fixTJunctions
- [ ] Convert to V2 math objects
- [ ] Rewrite to remove getTag
- [ ] Write test suite

retessellate (requires FuzzyFactory3d, see below)
- [X] rewite as part of Fix Fuzzy Factories #138 

transform
- [X] rewite as part of Fix Fuzzy Factories #138 

toCompactBinary (Changes to be completed with rewrite of fromCompactBinary)

### OPS CNC
getTransformationAndInverseTransformationToFlatLying
- [ ] Convert to V2 math objects
- [ ] Rewrite to remove getTag
- [ ] Write test suite

overCutInsideCorners
- [X] Convert to V2 math objects
- [X] Rewrite to remove getTag
- [ ] Write test suite IN PROGRESS

### GEOMETRY FuzzyFactory3
getPolygon
- [X] rewrite as part of Fix Fuzzy Factories #138


z3dev:
This is still incomplete but the logic is now using shape2 data structure for sides. The logic is easier to understand with the new data structures.

z3dev:
I have a few questions about shape2 outlineToPaths...
- This is only used by IO libraries. Would this be better as part of the IO library?
- This coverts sides to paths, which doesn't really make sense either. Wouldn't just a list of points be better?

pentacular:
One request :)

This is something that the original doesn't do, but might be easy to add at this point.

Could we maintain the path directions, such that the exteriors are CW, but holes are CCW?

This will allow a transformation from shape2 to paths to shape2, and should make conversion to winding depending outputs like svg easier to do.

It will also allow path-wise transformations of shape2.

If this isn't easy to do now, then I can follow up on it later.

PR #127, opened 2018-11-03
refactor(): Spliting of Shapes & geometry

kaosat-dev:
this PR splits out the geometry data & functions from the shapes:
- for seperation of concerns
- to enable us to swap out geometric engines if need be in the future
- to regroup all the very geometric specific code & data (details of geometric structure, flags like 'isRetesselated' etc) into a single, clearly identified object

### DIFF
--- a/src/core/connector/transformationBetweenConnectors.js
+++ b/src/core/connector/transformationBetweenConnectors.js
@@ -7,10 +7,10 @@ const transform = require('./transform')
 const normalize = require('./normalize')
 
 /** Get the transformation matrix to connect this Connector to another connector
- * @param {Object} params
+ * @param  {Object} params
  * @param  {Boolean} params.mirror=false the 'axis' vectors of the connectors should point in the same direction
  *  true: the 'axis' vectors of the connectors should point in opposite direction
- * @param {Number} params.normalrotation=0 : degrees of rotation between the 'normal' vectors of the two connectors
+ * @param  {Number} params.normalrotation=0 : degrees of rotation between the 'normal' vectors of the two connectors
  * @param  {Connector} from connector from which to connect
  * @param  {Connector} to connector to which the first connector should be connected
  */
@@ -54,6 +54,8 @@ const transformationBetweenConnectors = (params, from, to) => {
   transformation = mat4.multiply(transformation, mat4.fromTranslation(other.point))
   // let usAligned = us.transform(transformation);
   return transformation
+  // alternativeNaming ?
+  // mat4.fromConnectorPair(params, from, to)
 }
 
 module.exports = transformationBetweenConnectors


z3dev:
A few thoughts...

I'm more interested in seperating the API (that what users use) from the CORE (the guts and logic). And thought there would be some required 'translation' from API to CORE, for example circle ==> ellipse. I think this will happen naturally, as V2 takes shape, so will put this aside.

SHAPE vs GEOM

I see the advantages of seperating the SHAPE from the GEOM, which allows transforms/properites to be maintained seperated from the geometry. The ability to transform an object without recalculating geometry is a **huge benefit.**

It may be possible to swap out the geometry in the future as well, e.g. use large integers, etc. So making this split now may have some benefit in the future.

There are a few 'gray' areas to consider.
- Is there a 1-to-1 relation between SHAPE and GEOM? We need more than just shape2/shape3.
- When is GEOM require? Anything that holds 0:N math objects?
- Where does the logic reside? GEOM objects? SHAPE objects? Both? For example, the logic for shape2.difference and GEOM2.difference seems very simular.
- Should non-GEOM non-basic functions be pushed up to the API? For example, linearExtrude.

There's too much confusion in the various functions, api layers, etc. It's difficult to determin if the benefits of GEOM objects will be required in V2.

z3dev:
Continuing some thoughts...

It seems that the 'vtree' will also add some complexity to the managment of shapes, and possibly require some special functions for creating geometries when required. If so then the extra complexity of spliting SHAPE and GEOM may be unwise at this time.

z3dev:
Hmm... I'm starting to wonder about something. Let's say I have a bunch of shapes.

let a = cube({radius: [5,5,5]})
let b = translate([5,5,5], a)
let c = translate([10,10,10], b)
...

Now, I want to reuse those shapes to punch holes into another shape.

let x = rectangle({radius: 100,100,2})
let y = subtract(x, a, b, c) // this 'bakes in' the geometries, and performs the operation
y = rotate([0,0,90], y)
y = subtract(y, a, b, c) // this 'bakes in' the geometries, and performs the operation

My concern is the second subtract, where the geometries are 'baked' AGAIN.
Do we need a variable to determine if the state of the geometries, i.e. baked/unbaked? Do we reset the transform to the 'identity' matrix on 'baked' shapes?

kaosat-dev:
@z3dev Hm you are right that even with the V2 changes it is not perfect but you still have less baking than previously in total since I think there a lot more transforms than booleans in designs ? Otherwise it really is a big issue
Possible optimisations: 
- cached geometry (reuse of geometry with same transforms)
- ignore transform() operation on geometry if the transforms are the identity matrix (this seems the best)
- about  flag on geometries, hmm not sure how that would work, since the 'baked in' nature is specific to a given transform ?

The count of bakes is really not too good: 
#V1 :
let a = cube({radius: [5,5,5]})
let b = translate([5,5,5], a)  // bakes : 1 
let c = translate([10,10,10], b) // bakes 2 

let x = rectangle({radius: 100,100,2})
let y = subtract(x, a, b, c) // 
y = rotate([0,0,90], y) // bakes: 3
y = subtract(y, a, b, c)  //

# V2: 
let a = cube({radius: [5,5,5]})
let b = translate([5,5,5], a) 
let c = translate([10,10,10], b)

let x = rectangle({radius: 100,100,2})
let y = subtract(x, a, b, c) // bakes: 4
y = rotate([0,0,90], y) // 
y = subtract(y, a, b, c)  // bakes: 8: yikes !

z3dev:
@kaosat-dev thanks.

Agreed. We use the identity matrix to keep from baking again, or at all!

let a = cube({radius: [5,5,5]})
let b = translate([5,5,5], a)
let c = translate([10,10,10], b)

let x = rectangle({radius: 100,100,2})
let y = subtract(x, a, b, c) // bakes: 4
y = rotate([0,0,90], y) //
y = subtract(y, a, b, c) // bakes: 1

z3dev:
I understand that the transforms may be important for a design. However, I don’t think that a shape should be asked to supply the current transform applied. The transform (if important) should be managed as part of the design.

However, I should be able to measure the bounds, center, etc. Also, some additional measurements may be possible through other geometry representations.

Just thinking...

kaosat-dev:
Going to merge this one, although
- I still have strong doubts about the split between transforms & geometry
- it is not quite fully fleshed out (blame it on our still broken geometry)
- this still updates & overhauls a ton of things for V2

PR #130, opened 2018-12-15
fix center() for non-uniform geometries

jeremy-brenner:
Fixes this:

screen shot 2018-12-15 at 11 43 46 am


kaosat-dev:
@jeremy-brenner wow good catch ! Thanks for the fix !

PR #132, opened 2019-01-06
Verify all rotation functions across math objects

z3dev:
This small set of tests verify that the math objects (mat4, vec2, vec3) produce expected rotations, as well as produce the same results for simular functions.

P.S. Its a very confusing world out there about rotation, orientation, matrices, etc.

### DIFF
--- a/src/core/math/rotation.test.js
+++ b/src/core/math/rotation.test.js
@@ -1,0 +1,133 @@
+const test = require('ava')
+const { compareVectors } = require('../../../test/helpers/index')
+
+const {radToDeg, degToRad} = require('./utils')
+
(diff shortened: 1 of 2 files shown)


z3dev:
@kaosat-dev please review and merge. These tests are complete.

kaosat-dev:
@z3dev the tests are great, but I am still not 100% sure the premise is the right one
I 100% trust the math of gl-matrix, but not our own  :)

z3dev:
> @z3dev the tests are great, but I am still not 100% sure the premise is the right one
> I 100% trust the math of gl-matrix, but not our own :)

These are real simple tests to make sure WE don’t screw things up. As of now, all rotation functions are in sync. 😃

PR #134, opened 2019-01-16
add hull() for 3D solids implemented by using quickhull3d package

hg42:
* the function still works with 2d objects
* `chain_hull` works, too

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,9 @@
     "jscad"
   ],
   "license": "MIT",
-  "dependencies": {},
+  "dependencies": {
+    "quickhull3d": "^2.0.4"
+  },
   "devDependencies": {
     "ava": "^0.23.0",
     "conventional-changelog-cli": "^1.3.4",
--- a/src/api/ops-transformations.js
+++ b/src/api/ops-transformations.js
@@ -3,7 +3,9 @@ const Plane = require('../core/math/Plane')
 const Vector3 = require('../core/math/Vector3')
 const { union } = require('./ops-booleans')
 const { fromPoints } = require('../core/CAGFactories')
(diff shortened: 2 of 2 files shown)


hg42:
btw. I know this:
> as of September 1 2018, we will NOT be accepting new PRS against the V1 / master branch: we realistically
> are not able to keep adding new features/fixes to the 'old' version and work on the very different V2.
> thank you for your comprehension !

but it is currently not possible to test V2 with this, that's why I created this PR.

and sorry for force-pushing, I accidentally commited a wrong file, where 2d didn't work

z3dev:
@hg42 thanks for this pull request. The changes look spot on.

However, we have a problem... The quickhull3d package uses ES5 ‘import’ / ‘export’ which is good and proper, and the future. However, at this time, it seems that ‘import’ isn’t fully supported in browsers. This library needs to be available within web workers, which are used in the web console to compose and execute the designs.

If you have any more insights then please let us know. V2 is approaching, and support for 3D hull would be a great addition.

RedHatter:
This isn't a problem. Part of `quickhull3d`'s build process transpiles ES6 modules into require. Go ahead and install the package through npm. You'll find a dist folder not containing any ES6.

Browsers don't natively support require either. Both ES6 modules and require are typically run through a bundler before being ran on a client. Using `quickhull3d` won't change anything about how `csg.js` can be used or built.

z3dev:
@hg42 Thanks for this pull request.

I used this pull request to complete 3D hulls in V2.

If you don’t mind, can I close this pull request?

hg42:
sorry, I'm a bit offline lately...
thanks for integrating 3D hulls in V2

PR #138, opened 2019-01-24
Fix Fuzzy Factories

z3dev:
**WORK IN PROGRESS DO NOT MERGE**


kaosat-dev:
Hi @z3dev
Thanks for the awesome hard work ! :-) Will review in full over the weeekend.
Those linting commits are flat out wrong though :-( , then again linting has not been setup / updated correctly in this repo as far as I remember.

z3dev:
No problem about the lint settings. But the lint does find small issues, which is the most value. Please setup during your next pull request. Let’s start running lint continuously.

z3dev:
@kaosat-dev Thanks again.
I'll add the ESLINT configuration based on the IO library configuration for ESLINT. And, run lint again.
Should be done tomorow.

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -1,22 +1,24 @@
...
   this.planefactory = new FuzzyFactory(4, EPS)
   this.polygonsharedfactory = {}
 }
 
-FuzzyShape3Factory.prototype = {
+FuzzyFactory3.prototype = {
   getPolygonShared: function (sourceshared) {
-    let hash = polygonSHaredHash(sourceshared)
+    let hash = 'null'
+    if (sourceshared && (color in sourceshare)) {
```
pentacular: Having trouble seeing how this can work.

Where does color come from?
If sourceshared is falsey, does it share universally with 'null'.
Is 'null' ambiguous (perhaps with a 'null' color).

Suggest something closer to let hash = (sourceshare.color || []).join('/'); for now.

May also find that the following approach is a bit simpler:

let cached = this.polygonsharedfactory[hash];
if (cached === undefined) {
  cached = sourceshared;
  this.polygonsharedfactory[hash] = cached;
}
return cached;

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -48,27 +50,26 @@ FuzzyShape3Factory.prototype = {
...
-        if (vertextag !== prevvertextag) {
+        if (!vec3.equals(prevvertex, vertex)) {
           newverticesDedup.push(vertex)
         }
-        prevvertextag = vertextag
+        prevvertex = vertex
       })
     }
-    // If it's degenerate, remove all vertices:
+        // If it's degenerate, remove all vertices:
```
pentacular: May want a comment about why a degenerate polygon with no vertices is preferable to a degenerate polygon with 1 or 2 vertices. :)

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -48,27 +50,26 @@ FuzzyShape3Factory.prototype = {
     let newvertices = sourcepolygon.vertices.map(function (vertex) {
       return _this.getVertex(vertex)
     })
-    // two vertices that were originally very close may now have become
-    // truly identical (referring to the same Vertex object).
-    // Remove duplicate vertices:
-    // uses the tags to set vertices appart
+        // two vertices that were originally very close may now have become
```
pentacular: Is the indentation of the comment intentional?

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -26,15 +28,15 @@ FuzzyShape3Factory.prototype = {
...
     let result = this.vertexfactory.lookupOrCreate(elements, function (els) {
       return sourcevertex
     })
     return result
   },
 
   getPlane: function (sourceplane) {
-    let elements = [sourceplane.normal._x, sourceplane.normal._y, sourceplane.normal._z, sourceplane.w]
+    let elements = [sourceplane[0], sourceplane[1], sourceplane[2], sourceplane[3]]
     let result = this.planefactory.lookupOrCreate(elements, function (els) {
```
pentacular: Maybe
lookupOrCreate(elements, _ => sourceplane) might be clearer?

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -48,27 +50,26 @@ FuzzyShape3Factory.prototype = {
     let newvertices = sourcepolygon.vertices.map(function (vertex) {
```
pentacular: Maybe vertex => _this.getVertex(vertex)

src/core/geometry/fuzzyfactory/FuzzyFactory2d.js
```diff
@@ -1,11 +1,12 @@
...
-const FuzzyShape2Factory = function () {
+const FuzzyFactory2d = function () {
   this.vertexfactory = new FuzzyFactory(2, EPS)
 }
 
-FuzzyShape2Factory.prototype = {
+FuzzyFactory2d.prototype = {
   getVertex: function (sourcevertex) {
     let elements = [sourcevertex[0], sourcevertex[1]]
     // FIXME: oh blimey, this actually never did anything : returns the SOURCE VERTEX!!
```
pentacular: I think this comment is off -- it should only return sourcevertex if it can't find elements.

If sourcevertex is of the correct form then it can probably be use directly.

It might be clearer to write something like this.

lookupOrCreate(elements, otherwise => sourcevertex)

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -48,27 +50,26 @@ FuzzyShape3Factory.prototype = {
...
       return _this.getVertex(vertex)
     })
-    // two vertices that were originally very close may now have become
-    // truly identical (referring to the same Vertex object).
-    // Remove duplicate vertices:
-    // uses the tags to set vertices appart
+        // two vertices that were originally very close may now have become
+        // truly identical (referring to the same Vertex object).
+        // Remove duplicate vertices:
     let newverticesDedup = []
```
pentacular: Perhaps distinctVertices would be a clearer name?

src/core/geometry/fuzzyfactory/FuzzyFactory2d.js
z3dev: @kaosat-dev Should I move this code to moveOutOfCode? I don't think we will use this.

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
z3dev: Added a comment. Sadly, we haven't decided if color is built into poly3, or poly3 has a property list. TBD

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
```diff
@@ -26,15 +28,15 @@ FuzzyShape3Factory.prototype = {
...
+    let elements = [sourcevertex[0], sourcevertex[1], sourcevertex[2]]
     let result = this.vertexfactory.lookupOrCreate(elements, function (els) {
       return sourcevertex
     })
     return result
   },
 
   getPlane: function (sourceplane) {
-    let elements = [sourceplane.normal._x, sourceplane.normal._y, sourceplane.normal._z, sourceplane.w]
+    let elements = [sourceplane[0], sourceplane[1], sourceplane[2], sourceplane[3]]
```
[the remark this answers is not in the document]
z3dev: Yeah. It would be nice. I also thought of adding poly3.x() functions as well. But once you know that everything is stored as arrays in V2, then indexes 0, 1, 2, 3 just become natural.

src/core/geometry/fuzzyfactory/FuzzyFactory3.test.js
```diff
@@ -0,0 +1,36 @@
+const test = require('ava')
+
+const { EPS } = require('../../constants')
+const poly3 = require('../poly3')
+
+const FuzzyFactory3 = require('./FuzzyFactory3')
+
+const { comparePolygons } = require('../../../../test/helpers/index')
+
+test('FuzzyFactory3: factories for 3D geometries should function properly', (t) => {
```
kaosat-dev: @z3dev can we please clarify what 'function properly' means ? we should prefer some explicit things like `FuzzyFactory3: factories for 3D geometries should return a single polygon when merging 2 polygons`  etc

src/core/geometry/geom3/canonicalize.js
```diff
@@ -1,38 +1,37 @@
...
   }
-}
 
-const Geom3FromFuzzyFactory = (factory, sourcecsg) => {
-  let _this = factory
-  let newpolygons = []
-  sourcecsg.polygons.forEach(function (polygon) {
-    let newpolygon = _this.getPolygon(polygon)
-    // see getPolygon above: we may get a polygon with no vertices, discard it:
+  const factory = new FuzzyFactory3()
```
kaosat-dev: nice! way more concise  and readable

src/core/geometry/geom3/canonicalize.js
```diff
@@ -1,38 +1,37 @@
...
+
+  const newpolygons = []
+  geometry.polygons.forEach((polygon) => {
+    const newpolygon = factory.getPolygon(polygon)
+
+    // discard incomplete polygons
     if (newpolygon.vertices.length >= 3) {
```
kaosat-dev: should this discarding not take place in the factory itself ? this seems weirdly redundant ?
I would expect the factory to return valid things or at least pre-filtered so that we would not need to re-iterate here ?

src/core/geometry/geom3/canonicalize.test.js
```diff
@@ -0,0 +1,26 @@
+const test = require('ava')
+const { canonicalize, create, fromPoints } = require('./index')
+
+test('geom3: canonicalize() should return a geometry with proper values', (t) => {
```
kaosat-dev: same remark as above 'should return the same geometry or a new `

src/core/geometry/geom3/clone.js
```diff
@@ -5,9 +5,11 @@ const poly3 = require('../poly3')
  * @param  {Geom3} sourceGeometry the geometry to clone
  * @returns {Geom3} the new clone
  */
-const clone = sourceGeometry => {
+const clone = (sourceGeometry) => {
   return {
-    polygons: sourceGeometry.polygons.map(p => poly3.clone(p)),
+    polygons: sourceGeometry.polygons.map((p) => {
```
kaosat-dev: why the change here ? the previous was a perfectly valid syntax ?

src/core/geometry/geom3/difference.js
```diff
@@ -20,34 +21,29 @@ const Tree = require('../trees')
...
   a.invert()
   let result = fromPolygons(a.allPolygons())
-  if (doRetesselate) result = retesselate(result)
   if (doCanonicalize) result = canonicalize(result)
+  if (doRetesselate) result = retessellate(result)
+  return result
+}
+
+const subtract = (otherGeom3, ...geometries) => {
+  let result = otherGeom3
```
kaosat-dev: this will not work in case somebody passes a single item by accident, you need to
- add `toArray` to enforce always getting an array
- thrown an error if array length is 1

src/core/geometry/geom3/intersection.js
```diff
@@ -1,9 +1,9 @@
+const Tree = require('../trees')
+
 const fromPolygons = require('./fromPolygons')
-const retesselate = require('./retessellate')
+const retessellate = require('./retessellate')
```
kaosat-dev: oh wow nice catch on the typo :)

src/core/geometry/geom3/intersection.js
```diff
@@ -42,10 +33,20 @@ const intersectSub = (ohterCsg, geometry, doRetesselate, doCanonicalize) => {
...
   a.invert()
   let result = fromPolygons(a.allPolygons())
-  if (doRetesselate) result = retesselate(result)
   if (doCanonicalize) result = canonicalize(result)
+  if (doRetesselate) result = retessellate(result)
+  return result
+}
+
+const intersect = (otherGeom3, ...geometries) => {
+  let result = otherGeom3
```
kaosat-dev: same remark as for subtract, you need to ensure that geometries is an array !
as it is `...geometries` syntax unpacks an array of geometries (please take a look at rest/spread syntax for es6)

src/core/geometry/geom3/intersection.test.js
```diff
@@ -1,5 +1,5 @@
 const test = require('ava')
-const { intersection, create, fromPoints, toString } = require('./index')
+const { intersection, create, fromPoints } = require('./index')
```
kaosat-dev: yeaah no more strings :D

src/core/geometry/geom3/invert.js
```diff
@@ -8,10 +8,12 @@ const fromPolygons = require('./fromPolygons')
  * @example
  * let B = invert(A)
  */
-const invert = geometry => {
-  let flippedpolygons = geometry.polygons.map(polygon => poly3.flip(polygon))
-  return fromPolygons(flippedpolygons)
+const invert = (geometry) => {
+  const flippedpolygons = geometry.polygons.map((polygon) => {
```
kaosat-dev: same question as elsewhere : why the syntax change, the other is perferctly valid

src/core/geometry/geom3/reTesselateCoplanarPolygons.js
```diff
@@ -7,332 +7,334 @@ const poly3 = require('../poly3')
...
+  // TODO const shared = sourcepolygons[0].shared
+  const orthobasis = new OrthoNormalBasis(plane)
+  const polygonvertices2d = [] // array of array of Vector2D
+  const polygontopvertexindexes = [] // array of indexes of topmost vertex per polygon
+  const topy2polygonindexes = {}
+  const ycoordinatetopolygonindexes = {}
 
-    let xcoordinatebins = {}
-    let ycoordinatebins = {}
+  const ycoordinatebins = {}
```
kaosat-dev: the other bin was not used anymore right ?

src/core/geometry/geom3/reTesselateCoplanarPolygons.js
```diff
@@ -7,332 +7,334 @@ const poly3 = require('../poly3')
...
+  //        leftvertexindex: the index of the vertex at the left side of the polygon (lowest x)
+  //                         that is at or just above the current y coordinate
+  //        rightvertexindex: dito at right hand side of polygon
+  //        topleft, bottomleft: coordinates of the left side of the polygon crossing the current y coordinate
+  //        topright, bottomright: coordinates of the right hand side of the polygon crossing the current y coordinate
+  let activepolygons = []
+  let prevoutpolygonrow = []
+  for (let yindex = 0; yindex < ycoordinates.length; yindex++) {
+    const newoutpolygonrow = []
+    const ycoordinateasstring = ycoordinates[yindex]
```
kaosat-dev: so we still use strings ? :D

src/core/geometry/geom3/reTesselateCoplanarPolygons.js
```diff
@@ -7,332 +7,334 @@ const poly3 = require('../poly3')
...
+  //                      and polygonvertices2d arrays)
+  //        leftvertexindex: the index of the vertex at the left side of the polygon (lowest x)
+  //                         that is at or just above the current y coordinate
+  //        rightvertexindex: dito at right hand side of polygon
+  //        topleft, bottomleft: coordinates of the left side of the polygon crossing the current y coordinate
+  //        topright, bottomright: coordinates of the right hand side of the polygon crossing the current y coordinate
+  let activepolygons = []
+  let prevoutpolygonrow = []
+  for (let yindex = 0; yindex < ycoordinates.length; yindex++) {
+    const newoutpolygonrow = []
```
kaosat-dev: please switch to camel case for variables if possible 
ie `newoutpolygonrow` => `newOutPolygonRow`

src/core/geometry/geom3/reTesselateCoplanarPolygons.js
```diff
@@ -7,332 +7,334 @@ const poly3 = require('../poly3')
...
-            toprightvertexindex = i
-          }
-          let nextleftvertexindex = topleftvertexindex + 1
+    // update activepolygons for this y coordinate:
+    // - Remove any polygons that end at this y coordinate
+    // - update leftvertexindex and rightvertexindex (which point to the current vertex index
+    //   at the the left and right side of the polygon
+    // Iterate over all polygons that have a corner at this y coordinate:
+    const polygonindexeswithcorner = ycoordinatetopolygonindexes[ycoordinateasstring]
+    for (let activepolygonindex = 0; activepolygonindex < activepolygons.length; ++activepolygonindex) {
```
kaosat-dev: I seem to recall that extracting `activepolygons.length` to a variable before the loop is good optimising method

src/core/geometry/geom3/retessellate.js
```diff
@@ -10,45 +9,43 @@ const fromPolygons = require('./fromPolygons')
   Polygons are split at each sweep line, and the fragments are joined horizontally and vertically into larger polygons
   (making sure that we will end up with convex polygons).
 */
-const retessellate = geometry => {
+const retessellate = (geometry) => {
+  if (!geometry.isCanonicalized) {
+    throw new Error('geometry must be canonical, call canonicalize first')
```
kaosat-dev: nice !

kaosat-dev: hmm or perhaps it should canonicalize by itself ? what do you think @z3dev

src/core/geometry/geom3/retessellate.js
```diff
@@ -10,45 +9,43 @@ const fromPolygons = require('./fromPolygons')
...
-      }
-    }
-    let result = fromPolygons(destpolygons)
-    result.isRetesselated = true
-    // result = result.canonicalized();
-    result.properties = geometry.properties // keep original properties
-    return result
   }
+
+  const polygonsPerPlane = new Map()
```
kaosat-dev: very nice ! 
Stupid question , is a weakMap not better ? (I am quite ignorant)

src/core/geometry/geom3/retessellate.js
```diff
@@ -10,45 +9,43 @@ const fromPolygons = require('./fromPolygons')
...
+  geometry.polygons.forEach((polygon) => {
+    let values = polygonsPerPlane.get(polygon.plane)
+    if (values === undefined) {
+      values = [polygon]
+      polygonsPerPlane.set(polygon.plane, values)
```
kaosat-dev: nice !

src/core/geometry/geom3/retessellate.js
```diff
@@ -10,45 +9,43 @@ const fromPolygons = require('./fromPolygons')
...
+    } else {
+      const retesselayedpolygons = reTesselateCoplanarPolygons(sourcepolygons)
+      destpolygons = destpolygons.concat(retesselayedpolygons)
+    }
+  })
+
+  const result = fromPolygons(destpolygons)
+  result.isCanonicalized = geometry.isCanonicalized
+  result.isRetesselated = true
+  // TODO result.properties = geometry.properties // keep original properties
```
kaosat-dev: properties will be moved to the upper level (ie they are based on the transforms, not geometry, so geometries need not care :)

src/core/geometry/geom3/toString.js
```diff
@@ -1,10 +1,10 @@
 const poly3 = require('../poly3')
 
-const toString = geometry => {
+const toString = (geometry) => {
   return geometry.polygons.reduce((result, polygon) => {
     result += poly3.toString(polygon)
     return result
-  }, 'Geom3: ('+geometry.isCanonicalized+','+geometry.isRetesselated+')\n')
+  }, 'Geom3: (' + geometry.isCanonicalized + ',' + geometry.isRetesselated + ')\n')
```
kaosat-dev: oh my gad ! we forgot to switch to es6 template strings :D

src/core/geometry/geom3/toTriangles.js
```diff
@@ -4,17 +4,16 @@ const poly3 = require('../poly3')
  * @returns {Polygons} triangulated polygons
  */
 const toTriangles = (shape3) => {
-  let polygons = []
-  shape3.polygons.forEach(function (poly) {
-    let firstVertex = poly.vertices[0]
+  const polygons = []
+  shape3.polygons.forEach((poly) => {
```
kaosat-dev: nice

src/core/geometry/geom3/union.js
```diff
@@ -20,35 +20,34 @@ const unionForNonIntersecting = (otherGeom3, csg) => {
...
+  // FIXME: what to do with properties ????
+  // result.properties = otherGeom3.properties._merge(csg.properties)
+  if (doCanonicalize) result = canonicalize(result)
+  if (doRetesselate) result = retessellate(result)
+  return result
 }
 
 const union = (...solids) => {
-  let csgs = solids
+  const csgs = solids.slice()
```
kaosat-dev: this should be done like the other boolean operations (arrays etc)

src/core/math/OrthoNormalBasis.js
```diff
@@ -13,14 +15,14 @@ const Plane = require('./Plane')
 const OrthoNormalBasis = function (plane, rightvector) {
   if (arguments.length < 2) {
     // choose an arbitrary right hand vector, making sure it is somewhat orthogonal to the plane normal:
-    rightvector = plane.normal.randomNonParallelVector()
+    rightvector = vec3.random(plane)
   } else {
-    rightvector = new Vector3D(rightvector)
+    rightvector = rightvector
```
kaosat-dev: yay for simplicity of V2 :)

src/core/geometry/fuzzyfactory/FuzzyFactory2d.js
z3dev: I don't believe we will use FuzzyFactory2 so let's trash it early. :)

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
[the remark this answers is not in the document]
pentacular: What I am suggesting is array notation, just replacing the magic numbers with meaningful constants.

src/core/geometry/geom3/retessellate.js
z3dev: I'll add a note to the new issue for the fuzzy factory. The rewrite should make this even easier, possibly even merge the logic into retessellate.

src/core/geometry/geom3/canonicalize.js
z3dev: Added note to #139

src/core/geometry/geom3/retessellate.js
z3dev: Added comment to #139

src/core/geometry/geom3/retessellate.js
z3dev: weakMap doesn't allow iteration across the keys. it's limited to set(), has(), get() only. there's someothing about the state management that makes iteration impossible.

src/core/geometry/geom3/difference.js
z3dev: How about this? Trying to take adavantage of the rest functionality.

```
const subtract = (...geometries) => {
  let result = geometries.shift()
  ...
```

src/core/geometry/geom3/clone.js
z3dev: eslint warnings

i won't change back now because we haven't setup lint. there will be several rounds of cleanup ahead.

src/core/geometry/geom3/difference.js
z3dev: Things will blow up if less then two geometries are provided. But checking the length of geometries can also be performed. We should be consistent.

Which way is it? I prefer natural implosion at this level. Of course, the API version of differene() should be checking.

src/core/geometry/fuzzyfactory/FuzzyFactory2d.js
pentacular: Don't you need this to ensure unique vertices?

src/core/geometry/fuzzyfactory/FuzzyFactory3.js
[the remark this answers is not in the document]
pentacular: For the first three, maybe -- but not so sure about W ... :)

src/core/geometry/geom3/difference.js
pentacular: Why not consider subtracting no geometries as a no-op?

src/core/geometry/geom3/intersection.js
pentacular: Are you sure?

Perhaps I've misunderstood, but I don't think that ...geometries will unpack a passed array, but rather will pack a local array.

e.g., ((...x) => x)([1, 2]) produces [ [ 1, 2 ] ] for me.

The [1, 2] isn't unpacked, but is rather packed into an array as [[1, 2]]

src/core/geometry/geom3/union.js
z3dev: I added tests for zero or one geometry passed to difference, intersection, and union. All behave the same now, and return undefined when passed zero geomeries, as well as returning the original geometry if passed one geometry.

pentacular: Wouldn't the empty geometry be correct for no arguments?

Returning undefined sometimes means that you need to either guard the inputs or check the outputs.

z3dev:
@kaosat-dev pleae review difference, insection, and union again.

i made changes to flatten the parameters to arrays. and added a check for zero parameters as well, which now throws an error.

kaosat-dev:
@z3dev awesome ! To me this is all good ! shall we merge ? or do you have other changes planned ?

PR #141, opened 2019-02-02
Fix chainHull

pentacular:
Fixes shape2/chainHull and adds test.

hullChain is removed from geom2, due to not being primitive.
chainHull is redefined in terms of hull and union, and given es6 interface.

Introduces rewiremock for module level mocking.
Adds eslint dev dependency and configuration.

Some strangeness in the test, and there is probably a better approach -- ptal.


kaosat-dev:
Hi @pentacular  , thanks for the contribution ! 
Very sorry , but there are multiple things that feel off about this to be honest:
- addition of esLint config: good, very useful, we need this thanks !
- I honestly do not understand or agree with some of your changes to `chainHull`: 
    * handling of 'closed' seems good :) 
    * it is working on geometries not shapes, hence why it was in that place
    * the name was changed to `hullChain`for a good reason :) 
    * can you explain why you added all the mocking setup ? (just to be clear this is not a passive aggressive question , but an actual one :)

z3dev:
This actually asks one of the questions rolling around in my head.

We want to support operations which take multiple objects, like hullChain. But does each layer need to support the same function call?

API... chainHull(...objects)
SHAPE... chainHull(...objects)
GEOM... chainHull(...objects)

It seems to me that API and SHAPE must, as SHAPE provides the specialized functionality to create GEOM. But I don’t think that operations between GEOM objects needs to deal with multiple objects, and could be more basic.

GEOM... hull(...)

z3dev:
By the way... isn’t hull(...shapes) the same as hullChain(...shapes)? Shouldn’t we just have hull() in the API?

pentacular:
Please note that I don't have any particularly strong opinions about this change or operation.
I picked it because it seemed trivial and was the first thing that broke. :)

My main goal here is to produce a consensus that will work going forward generally.
So I'm happy to discuss at length how things ought to be in a principled fashion.
And I expect some of my expectations and understandings may need adjusting.

My understanding and reasoning is as follows:

(a) A shape is { geom + matrix }.

Hulling needs to consider orientation and position, which geom lacks.
Therefore hulling must be a shape operation, rather than a geom operation.

You can certainly implement that shape operation by creating a transformed geom and operating on that, but since this operation is essentially a combination of union and hull, it seemed more straight-forward to compose those primitive operations instead of adding a new geom primitive.

I think also that this operation should probably move out of the shape2 api and into a utility library, at some point, due to not being primitive.

(b) chainHull vs hullChain

Actually, I didn't change any names here :)

I just deleted the one in geom, and modified the existing implementation in shape2.

I understand you're using verbNoun form, so I guessed it should be chainHull due to chaining (verb) hulls (noun) -- but I'm happy to rename it if you like.

(c) Module Mocking

I added module mocking because otherwise we're testing implementation details. :)

Which is particularly critical at the moment, because those implementation details (like union) are broken, so using mocks allows us to test this part while the other parts are broken.

But in general we probably shouldn't be testing those other parts anyhow.

It's the first time that I've used rewiremock, and I haven't used ava significantly, so there may be better choices and approaches, but I do not see how you'll be able to do principled testing under the current design without module level mocking.

If you have a better way to do it, then I'd be happy to change.

(d) APIs

Regarding APIs, I guess it depends on what will be external and what will be internal.

If end-users should call geom interfaces, then having packed and curryable interface there makes sense.

If end-users should not call geom interfaces, then probably they should receive an unpacked interface that does not care about currying (unless there's a reason to curry geom calls).

In which case you might expect

shape.chainHull({ closed: true }, a, b, c)
but
geom.chainHull([a, b, c], true)

Or you might say that shape isn't an external api, and introduce an external layer elsewhere to do the user friendly interface unpacking, and support currying --- a bit like the current scad api module.

(I assume you only care about currying in the external layer to support pipe-lining?)

z3dev:
> Regarding APIs, I guess it depends on what will be external and what will be internal.
> 
> If end-users should call geom interfaces, then having packed and curryable interface there makes sense.

The geometry objects should be for internal use only. Users may want to use the shapes directly if making something new, like star shapes. 

> If end-users should not call geom interfaces, then probably they should receive an unpacked interface that does not care about currying (unless there's a reason to curry geom calls).

A very good idea. I think geometry objects should have functional interfaces that require full options, nothing is default, nothing is optional. The API and SHAPE logic should be making all the guesses and translations.

> Or you might say that shape isn't an external api, and introduce an external layer elsewhere to do the user friendly interface unpacking, and support currying --- a bit like the current scad api module.

I’m thinking that another API should be possible as well. We have users interested in building other paradigms, so we should expect other uses. The difficulty here ... what should be core and what should be API?

kaosat-dev:
Just 2 cents before I reply to the rest 
> The geometry objects should be for internal use only. Users may want to use the shapes directly if making something new, like star shapes.
I completely dissagree on this. We should just provide useful building blocks, there is no need to 'hide' anything: if a user just wants to manipulate geometry, then it is fine : we can impose having to wrap your final results in your scripts in a shape, but other than that I think it is not a good idea to artificially close off things or to impose use of Shapes where geometries suffice

pentacular:
What I wonder is -- what's the real difference between a geom and a shape?

Is it just that a shape has lazy transforms, and a geom has eager transforms?
If so, why not merge the two, and have lazy transforms which can be forced?

I guess there's a good reason, but the current interfaces don't make it very obvious to me.

kaosat-dev:
@pentacular : 
- a shape , as you put it , is 
 * geometry + transforms: 250 copies of the same shape with different transforms should not need to duplicate the geometry from scratch, it is also essential to have a seperate transform to be able to create scene graphs etc
- a geometry is 'just' the actual structure of an object, its topology

We discussed merging them vs keeping them seperate, and the problem is , they are not dealing with the same things, so you end up conflating concerns, and end up with something even messier

pentacular:
I think there's something missing there.

Either a shape is a 'geometry with lazy transforms', in which case they're dealing with the same things -- the only difference is in performance.

Or they are different kinds of things dealing with different concerns, in which case a shape is not a 'geometry with lazy transforms'.

What's confusing me is that what you've written above makes both claims. :)

kaosat-dev:
I honestly do not know what more to tell you , the seperation between those concerns (and having actual information about the current transforms, and not just 'baked into' the vertices) is a very standard practice in most 3d tools.
'geometry with lazy transforms' is only partially accurate: if you mean 'baking in' the transform into the geometry that is true, but that is not always needed !
You cannot create any remotely useable hierarchies of  objects / transforms (parent <> children) without having geometries seperate from transforms

kaosat-dev:
as to 'why not have a single object that has both'
multiple (semi similar) reasons :
- seperation of concerns
- specialization of tools (for example you could use the new geometry format relatively easilly with some of the modules of the stack.gl ecosystem, it would be a bit more complicated if they where conflated)
- transforms structure might change (ie translation + rotation as quaternion might be better ? )

pentacular:
Sure, which brings me back to the question.

Why not just add lazy transforms to shape and not bother with a separate geom?

kaosat-dev:
I do not get your question, sorry :(

kaosat-dev:
shape transforms are already lazy ?
https://github.com/jscad/csg.js/blob/V2/src/core/shape2/transform.js#L11

pentacular:
Let's say geom and shape weren't separated -- there was just shape2 and shape3.

Then let's say we added a matrix property to shape2 and shape3.
And then had the transforms update the matrix instead of the sides or polygons.

And then, when we needed to combine objects with different accumulated transforms (i.e., s1.matrix != s2.matrix) we transformed the sides or polygons so that the transforms were compatible (e.g., so that s1.matrix == s2.matrix == identityMatrix).

Then we would have the same benefits, but there wouldn't be an idea of geom2 or geom3.

--

So I guess the next step would be to implement union, intersection, difference, transform on sides and polygons.

And then arrange them into separate objects which you could call geom2 and geom3.

--

But we end up with a lot of mess in geom2 and geom3 -- things like extrude, and hull, and measureBounds, and so on, and a lot of duplicated interfaces between shape and geom.

So maybe it should have gone the other way -- separating out the lazy operations with a matrix object, and putting that alongside the geometry.

Now we end up with no duplicated interfaces, and just have the transform operations update the matrix, and an operation like applyTransforms() to apply those operations to the sides and polygons (if it doesn't already have an identity transform).

--

I guess you've already considered this approach -- it would help to understand why it was rejected.

That is, why not get rid of shape, and instead add lazy transform support to geom?

Issue #139, opened 2019-01-28
rewite fuzzy factories


z3dev:
Fuzzy Factories are really inefficient, and could be a cause of issues in solids.

The current versions of the fuzzy factories:
- converts numbers to strings for hash values '99999/99999/9999'
- creates two (2) hash values to 'correct' rounding issues
- uses object attributes for storing points by hash values
- looks up stored points by hash, which is basically just full traversals
- uses EPS as a numeric tolerence, does really compare distances between points
- uses the same logic for looking up planes, which is just WRONG due to the precision required for normals
- mixes planes (normals) and points from unrelated polygons, which could lead to all kinds of issues

REWORK Option 1 : Reuse Points within EPS distance
- rework the logic to use a simple map of points
- compare the distance between points
- reuse points from the map if distance < EPS
- if a point is reused then always calculate a new plane

REWORK Option 2 : Round All Points
- define a tolerence for all points
- round all points to the tolerence
- always calcuate a new plane

kaosat-dev:
Great clear points @z3dev !
Is this a list of ideas options AFTER your current PR fixing things ? or complimentary ?

z3dev:
I didn’t change anything in the PR #138 . I didn’t want to break everything.
This issue is for later persual.

z3dev:
Notes from reviews:
- geom3/cannonalize will also need changes, possibly elimination if possible

pentacular:
Definitely option 2, else you'll end up with accumulated error.

I was just thinking of doing that to make geom2 tests reasonable.

DanielJoyce:
Cannonalize is a typo. The real word is canonicalize.

z3dev:
V2 eliminates fuzzy factories.

Closing this issue.

PR #144, opened 2019-02-05
Vec2

pentacular:
Canonicalization via EPS based deduplication leads to incremental creeping error and can bring geometries out of plane.

Quantization of space avoids this issue.

This adds support for a quantization of space to a resolution of 100 nm, which equals the current EPS threshold.

Initial support is added to vec2 via a canonicalization operation which rounds the vector components to the nearest 100 quantized value.

### DIFF
--- a/src/core/math/vec2/canonicalize.js
+++ b/src/core/math/vec2/canonicalize.js
@@ -1,0 +1,8 @@
+const spatialResolution = require('../../constants').EPS;
+const fromValues = require('./fromValues');
+
+const quantize = (value) => (Math.round(value * spatialResolution) / spatialResolution)
+
+const canonicalize = (vector) => fromValues(quantize(vector[0]), quantize(vector[1]))
+
+module.exports = canonicalize
--- a/src/core/math/vec2/canonicalize.js
+++ b/src/core/math/vec2/canonicalize.js
@@ -1,4 +1,4 @@
-const spatialResolution = require('../../constants').EPS;
+const { spatialResolution } = require('../../constants')
 const fromValues = require('./fromValues');
 
 const quantize = (value) => (Math.round(value * spatialResolution) / spatialResolution)
--- a/src/core/math/vec2/canonicalize.test.js
+++ b/src/core/math/vec2/canonicalize.test.js
@@ -1,10 +1,15 @@
 const canonicalize = require('./canonicalize')
 const equals = require('./equals')
 const fromValues = require('./fromValues')
-const spatialResolution = require('../../constants').spatialResolution
 const test = require('ava')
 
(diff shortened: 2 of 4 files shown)


src/core/math/vec2/canonicalize.js
```diff
@@ -0,0 +1,8 @@
+const { spatialResolution } = require('../../constants')
+const fromValues = require('./fromValues');
+
+const quantize = (value) => (Math.round(value * spatialResolution) / spatialResolution)
```
kaosat-dev: very nice ! simple and to the point !

z3dev:
Does this partiall address the current issues in the fuzzy factories? #139

src/core/math/vec2/canonicalize.test.js
```diff
@@ -0,0 +1,20 @@
+const canonicalize = require('./canonicalize')
+const equals = require('./equals')
+const fromValues = require('./fromValues')
+const test = require('ava')
+
+const piVec = fromValues(3.141592653589793, 3.141592653589793)
+
+// This test is intended to be illustrative, and establish ground truth
+// It will need to be updated if ../../constants.spatialResolution changes.
+test('vec2: Canonicalizion quantizes to spatial resolution', t => {
```
kaosat-dev: test is great except for spelling :) 
- canonicalization instead of Canonicalizion
- idempotent instead of idemponent

pentacular:
I hope so. :)

At least once geometries canonicalize their points when being canonicalized

src/core/math/vec2/canonicalize.test.js
pentacular: Thanks for spotting those.

I have a little dyslexia, which makes it hard to see those in long words. :)

kaosat-dev: no worries, I had to double check as well :)

kaosat-dev:
@z3dev I think this is fine to merge, shall I ?
@pentacular thanks for this addition :)

kaosat-dev:
small note @pentacular I think it might make sense in the future to not use a constant but rather passing the value explicitly (or perhaps have a constant BUT pass it in explicitly)
We had issues in the past with constants and the fact that some of them are fairly arbitrary and should actually be parameters :) 
(just my 2 cents, I do not think it matters for this PR)

pentacular:
In general, I would agree, but I suspect that having multiple notions of spatial quantization in the same system at the same time is just asking for trouble.

Making all of those constants into parameters of the module might make sense in the future.

z3dev:
I agree. Users should be able to set the tolerances / epsilon values, but there’s a relationship between most, so a function to set a basic tolerance, maybe distanceEPS, could also set the other values.

PR #145, opened 2019-02-06
Fix geom3 measurements

z3dev:
This is a small pull request to complete the conversions of the geom3 mesaurement functions to V2 math, correct issues (there were some), add documentation, and add tests.

Also, some improved documention and tests for poly3/measureSignedVolume


src/core/geometry/geom3/measureArea.js
```diff
@@ -1,9 +1,13 @@
...
-  let result = toTriangles(geometry).map(triPoly => poly3.measureArea(triPoly))
-  console.log('area', result)
+/*
+ * Measure the area of the given geometry.
+ *
+ * @param {Geom3} geometry - 3D geometry to measure
+ * @returns {Number} area of the geometry
+ */
+const measureArea = (geometry) => {
+  let result = geometry.polygons.reduce((area, polygon) => area + poly3.measureArea(polygon), 0)
```
kaosat-dev: oh wow nice catch !

src/core/geometry/geom3/measureBounds.js
```diff
@@ -2,34 +2,25 @@ const vec3 = require('../../math/vec3')
...
+      vec3.clone(minpoint, bounds[0])
+      vec3.clone(maxpoint, bounds[1])
+    } else {
+      vec3.min(minpoint, minpoint, bounds[0])
+      vec3.max(maxpoint, maxpoint, bounds[1])
     }
-    // FIXME: not ideal, we are mutating the input, we need to move some of it out
-    geometry.cachedBoundingBox = [minpoint, maxpoint]
   }
-  return geometry.cachedBoundingBox
```
kaosat-dev: we might still need caching of bounds ...quite computationally expensive
(hence storing them along side geometry & transforms, as we discussed)

src/core/geometry/geom3/measureVolume.js
```diff
@@ -1,9 +1,13 @@
...
-  let result = toTriangles(geometry).map(triPoly => poly3.getSignedVolume(triPoly))
-  console.log('volume', result)
+/*
+ * Measure the volume of the given geometry.
+ *
+ * @param {Geom3} geometry - 3D geometry to measure
+ * @returns {Number} volume of the geometry
+ */
+const measureVolume = (geometry) => {
+  let result = geometry.polygons.reduce((volume, polygon) => volume + poly3.measureSignedVolume(polygon), 0)
```
kaosat-dev: why the signed volume?

src/core/geometry/poly3/measureSignedVolume.js
```diff
@@ -1,9 +1,15 @@
 const vec3 = require('../../math/vec3')
 
-// NOTE: This is totally worthless. The volume of a planar polygon is always zero(0).
+/*
```
kaosat-dev: great job adding docstrings !

src/core/geometry/geom3/measureBounds.js
z3dev: Agreed. We can cache at the appropriate level. So far, we have...
- properties (values that can be transformed)
- measurements (cached calculations)
- connectors (??)

src/core/geometry/geom3/measureVolume.js
z3dev: I tried to test volumes in ’negative’ space, but found negative - negative produced positive volumes.

Any thoughts?

kaosat-dev: good point, is a volume not always positive ? or does the sign have a significance ?

z3dev: It seems to be positive always. The research paper seems to suggest that negative volumes are possible, but I cannot produce a test case.

PR #147, opened 2019-02-06
A new surface geometry (independent of geom3 / CSG)

pentacular:
This is a surface geometry supporting union, intersection, difference, transform, etc.

It supplies a minimal set of primitives, and interoperates via polygon arrays, making it completely decoupled from other geometries (other than vec3).

Lazy in-place transform is supported.
The internal representation is based on polygons represented as paths, rather than as a set of sides.
The heavy lifting is delegated to the polygon-clipping module, which has only one dependency.

Operations that lie between geometries (e.g., extrude, hull, etc) have been removed and implementations based on polygon and/or point array transforms are planned to follow.

Non-primitive operations, such as rotateX have been removed and are planned to move to a generic user API.

Please see the following document for design philosophy, and feel free to comment or suggest changes.

https://docs.google.com/document/d/1nz7dhtErHBqBpuFcu7ooRU91miVTD4UxiJ6od1TbUgs/edit#heading=h.vyemavepdkpd


z3dev:
Point #1....

I’m not opposed to replacing the BSP (?) tree with a library, but we should talk about it. We know that the current operations have issues.

Do you have any concerns with the polygon-clipping module? I noticed that you wanted to discuss with the owner... what are the discussion points?

pentacular:
What I am proposing here is not a replacement, per se, but an alternative implementation.
This should be able to coexist with geom2, assuming geom2 supports a compatible interface.

Once we have a standard interface for surface geometries, we can easily perform tagged dispatch to select between them on a per-geometry basis.

e.g., instead of writing geom2.union(a, b) or surf2pc we could write something like implementationOf(a).union(a, b) which would find a tag on the geometry and map it to the appropriate module for its implementation. This would preserve current semantics. (And as always, we can bury it out of sight in the user api)

And given interchange representations, like polygon array, we could automatically convert from geom2 to surf2pc or vice verse, although it would not be free.

I don't have anything to discuss with the polygon-clipping owners at the moment -- we'll need to run it through a large test suite to properly evaluate it.

If it doesn't work well, then we can make a surf2foo or a surf2bar to try out instead -- the heavy lifting is in the algorithms, rather than the glue.

z3dev:
Thanks. I understood that from the document. The interchange of implementation is also one of the goals, but not the main goal for V2. Again, I think this discussion moves things towards the right direction.

kaosat-dev:
- 
> Operations that lie between geometries (e.g., extrude, hull, etc) have been removed and implementations based on polygon and/or point array transforms are planned to follow.
I agree (cross use of existing structures) & can understand this in a way, but ***geom2 IS a points array*** as it is, so why remove it ?

- use of an external library : that is a flat 'no' at this stage sorry, alternative implementations and external module use is going to be ok , but only after V2 

- very important and sorry if I missed this, but can you do 'holes' / compound shapes with this ?

pentacular:
> I agree (cross use of existing structures) & can understand this in a way, but _**geom2 IS a points array**_ as it is, so why remove it ?

geom2 isn't a points array.
It is an unordered set of sides, wrapped up in an abstraction.

Also, my proposal is to remove 'Operations that lie between geometries (e.g., extrude, hull, etc)'
So I'm not sure why you're talking about removing geom2.

> * use of an external library : that is a flat 'no' at this stage sorry, alternative implementations and external module use is going to be ok , but only after V2

Can you provide reasoning for this 'flat no'?

> * very important and sorry if I missed this, but can you do 'holes' / compound shapes with this ?

Yes.

PR #149, opened 2019-02-08
Simple tagged dispatch for alternative and external geometry implementation.

pentacular:
This allows alternative or externally defined geometries to be used with the jscad api.

A geometry module should add something like.

const registerImplementation = require('.../registerImplementation')
const __jscadTag__ = 'some-unique-identifier'

registerImplementation(__jscadTag__, module)

And produce objects which carry { __jscadTag__: __jscadTag }.

Once that module is imported toImplementation(object) will produce that module, allowing constructions like.

toImplementation(object).eachPoint({}, fun, object)

See design in:
https://docs.google.com/document/d/1nz7dhtErHBqBpuFcu7ooRU91miVTD4UxiJ6od1TbUgs/edit#

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -1,32 +1,6 @@
-
-const primitives3d = require('./primitives3d-api')
-const primitives2d = require('./primitives2d-api')
-const booleanOps = require('./booleans/ops-booleans')
-const transformations = require('./transformations/ops-transformations')
-const extrusions = require('./ops-extrusions')
-const color = require('./color')
-const maths = require('./maths')
-const text = require('./text')
-const { echo } = require('./debug')
-
-// these are 'external' to this api and we basically just re-export for old api compatibility
-// ...needs to be reviewed
-const { CAG, CSG } = require('../../csg')
-
 const exportedApi = {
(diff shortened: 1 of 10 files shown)


kaosat-dev:
hmm I like the concept behind this (as we discussed having multiple implementations is something we want to enable for the future), but I am not sure about the implementation at all
- why do you need to 'register anything at all ? 
```
const bspGeom  = require('bspGeom')
const hedgeGeom = require('halfEdgeGeom')

hedgeGeom.eachPoint(object, fun, object)
// OR
bspGeom.eachPoint(object, fun, object)
// use which ever you need, if they provide same api of course
```

pentacular:
In your example you've tightly coupled the operation to the geometry.
This is reasonable for a function in the geometry interface.

But it is unworkable for functions in the user api.

Consider a translate convenience function, for example.
 
const translate = (vector, shape) => transform(fromTranslation(vector), shape)

But which 'transform' function do we need?

const translate = (vector, shape) => toImplementation(shape).transform(fromTranslation(vector), shape)

Now, the user apis are decoupled from the geometries and can be extended independently.
e.g., by users.

PR #150, opened 2019-02-08
Conversion function for polygon array to STL ascii string.

pentacular:
Given that the interface to @jscad/io is broken, there doesn't seem to be much benefit in keeping it around.

Here is an initial refactoring of the STL (ascii) serializer, which will be sufficient to start testing complex compositions in the real world. The test example has been verified as correct and manifold by slic3r.

This initial implementation is entirely decoupled from geometry representation by using an interchange representation as input.

The remaining conversion functions can be refactored across piece by piece from @jscad/io, and conversions, such as DST and PDF can be added.

Due to the decoupled design, conversion functions can also be moved out into separate packages or provided by users.

See design here:
https://docs.google.com/document/d/1nz7dhtErHBqBpuFcu7ooRU91miVTD4UxiJ6od1TbUgs/edit#


kaosat-dev:
Hi @pentacular, thanks for all your contributions , but  please
- take a bit more time on each PR, and actually interacting with us
- leave us more time to actually evaluate your PR
- take into account that we might not agree with all your proposals
- please take a closer look at what currently is in place and WHY 
- serializers and deserializers will be tackled AFTER V2 basics for csg.js are done, tackling them now is premature
- serializers and deserializers are not part of this repository , and each format has its own repository
- work in progress V2 does not mean V1 is broken: the packages are published and the whole system & ecosystem of modules are living on

pentacular:
Where have I not been actively responding to comments?

Where have I placed a time limit on the evaluation of a PR?

Where have I not been actively asking for discussion and feedback?

How do you expect to be able to actually do end-to-end tests of V2 without being able to produce real output?

z3dev:
@pentacular I think you would be far happier with ONE BIG MONO repository, which brings all the pieces together then the pieces can be work with together. you already know this is work in progress as well.

end-to-end comes when we have end to end ready. i don't think we are very close yet.

for testing this library, we can add additional dependencies for testing. wouldn't JSON output be more useful to test suites, external libraries, etc?

pentacular:
I think that I would really be happier with many small repositories -- math, each geometry, various algorithms, generators, serializers, etc, can all be decoupled.

And I'll work toward this in the longer term.

But the next best thing (at least in the interim) would be a single repository so that all versions are properly sync'd.

If that's not doable, then I'll accelerate the process of breaking things up.

--

I'm running end-to-end at the moment (which is what I wrote this PR up to be able to do).

My current end-to-end test uses a regularPolygon generator function, and a linear extrude function, then assembles geometries for translate and union, and then uses the function here to render the result as STL.

There's a little debugging work to do, and a fair bit of user api work to do, but it's not far off being a working system for me.

z3dev:
You didn’t answer my question on JSON. Isn’t that a better format for debugging as well as test cases?

pentacular:
Assuming you're talking about the JSON serializer in @jscad/io, I don't think so.

It's just a dump of the CSG internals, and not renderable by any independent tools, which makes it a poor source of ground truth.

You could render it using V1, but since the V2 code is derived from V1, if you use it as a baseline you'll find any bugs inherited from V1 to be invisible -- since they're part of your baseline. :)

On the other hand, STL can be rendered by a whole bunch of visualization tools, and there are many defect analyzers and so on, written by many people, which gives it a good grounding in reality.

If not, please let me know what you mean by JSON here. :)

kaosat-dev:
> I think that I would really be happier with many small repositories -- math, each geometry, various algorithms, generators, serializers, etc, can all be decoupled.

https://github.com/jscad/io/tree/V2/packages
https://github.com/jscad/OpenJSCAD.org/tree/master/packages

I guess it is our fault for not making it clear that the small specialized modules is already something **we** are doing & will be increasingly doing.

seing your answer above you seem to be aware of Io though :)

pentacular:
Glad to hear it -- any timeline for breaking up @jscad/csg?

kaosat-dev:
whenever it makes sense and is possible, but please let us tackle that, as there are lots of gotchas and pre-conditions.

kaosat-dev:
also fyi the first ones to be split out will be maths & colors (because they are comparatively simple & stable)

PR #151, opened 2019-02-09
V2 generations of 2D and 3D primitives

z3dev:
WORK IN PROGRESS DO NOT MERGE

As discussed, here are the generation functions for the 2D and 3D primitives as converted to V2 geometry and math. Basically, the functions were created from the logic as found in the API layer, api/primitive directory. All the functions work fine, as I have verified that the solids produced are correct.

However... This is just the start of the reorganization, so some additional changes are expected.

1.  The use of shape3 is mandatory. Are we going to keep the shape / geom split or merge? NO
2.  How's the split of functionality? Does 'primitives' make sense. As a first try, this seemed sensible.
3. There's a lot of tests that need to be completed. These are difficult to produce, so any suggestions would be appreciated.

Changes Required:
- [ ] Add simple checks of parameters; completeness, toxic combinations, etc. (API does type checks and/or conversions)

Test Suites To Be Complete:
- [ ] Add tests for all variations of options
- [ ] Add tests for error conditions


pentacular:
The concern that I have here is that these are tightly coupled with shape2 and shape3.

Could we change the generators to produce either (a) arrays of points, or (b) arrays of polygons.

And then compose those with calls to shape2.fromPoints, etc, separately?

Then we could use the circle generator to build a path instead, or run it through a post-processing function to do a novel kind of extrusion, or whatever else we like.

pentacular:
The other thing that I observe is that most of these aren't primitive, and could be easily decomposed into simpler forms using extrude, hull, etc.

So it might be that these belong in the user api, rather than core.

z3dev:
> The other thing that I observe is that most of these aren't primitive, and could be easily decomposed into simpler forms using extrude, hull, etc.
> 
> So it might be that these belong in the user api, rather than core.

All of those options would add performance degradation. Hopefully, the primitives are exactly that...
- pure math
- correct geometry

z3dev:
I have a different pull request staged and ready. So, this pull request is not required.

PR #153, opened 2019-02-10
math/line3 is broken

pentacular:
math/line3 seems to be confused between lines and vectors on lines and line segments.

In particular line3 equals is wrong -- two mathematically equal lines will generally compare false.

I think we need to decide what line3 is trying to describe before fixing it.

### DIFF
--- a/src/core/math/line3/equals.js
+++ b/src/core/math/line3/equals.js
@@ -3,13 +3,16 @@ const vec3 = require('../vec3')
 /**
  * Compare the given 3D lines for equality
  *
+ * FIX: This does not compare lines for equality.
+ *
  * @return {boolean} true if lines are equal
  */
 const equals = (line1, line2) => {
   // compare directions (unit vectors)
   if (!vec3.equals(line1[1], line2[1])) return false
 
-  // compare points
+  // FIX: This is wrong -- it will compare false for any two distinct points
+  // on the line.
   if (!vec3.equals(line1[0], line2[0])) return false
 
   // why would lines with the same slope (direction) and different points be equal?


z3dev:
Actually, I don’t understand the existence of these two objects, line2 and line3. These seem to be unbounded lines with slope. It seems that line2 is supposed to be a line defined in Hesse normal form. I thought line3 is suppose to be to complement in 3D.

line2 is used.
line3 is not used.

kaosat-dev:
agreed, we should find a way to remove it if not absolutely needed:
Here is the listing of where it was/is used:
- connectors : removed in V2
- orthoBasis: still in use: but you had a great idea to replace this with simpler things as per your document @pentacular

pentacular:
Actually, it's been very useful in the watertightness detection code I've put together before trying to replace fixTJunctions.

I think that I understand what line3 is now, and I think it mostly needs documentation and some changes to the equals operator.

They are a combination of (point, and unit vector) to describe a line, which is valid.

The problem is in the semantics of equals -- what is line3.equals measuring?

Currently it measures the equality of the implementation as (point, and unit vector), but doesn't measure line equality.

Elsewhere I've run into some other problems with equals, where it isn't practically possible to compute total equality, but is practical to compute 'definite inequality'.

For example, in the geometries, there are many possible isomorphisms.

I'll think about it some more --- it might just be that equals needs to be a little more nuanced.

z3dev:
Maybe we shouldn’t have equal which implies exact equality. Maybe a better check is equivalence between lines... meaning the two lines represent the same line. S, the check would include...
- distance between lines < EPS
- normals are near equal

equivalent(a, b)

pentacular:
I saw some suggestion that line3 be called a 'ray' somewhere, which sounds about right, as it's direction from a point.

So, how about line3 -> ray3 and worry about the rest later?

z3dev:
It’s not a Ray in the true sense, so kept Line. The functionality also works like an line, as you can see in the test cases.

https://en.m.wikipedia.org/wiki/Line_(geometry)

Also, shouldn’t this just be an issue? There isn’t any code change.

pentacular:
> It’s not a Ray in the true sense, so kept Line. The functionality also works like an line, as you can see in the test cases.

Except that they can't be compared as lines (as far as I can see).

> 
> https://en.m.wikipedia.org/wiki/Line_(geometry)
> 
> Also, shouldn’t this just be an issue? There isn’t any code change.

Probably. :)

PR #155, opened 2019-02-11
V2 : Split of math objects from core, in preparation of packaging

z3dev:
This pull requests moves math from src/core/math to src/math, which will allow the math objects to be used outside of the geometry of the CSG library. The expectation is that the math objects will be packaged and published as a small reusable component.

FYI. **This pull request will NOT pass all checks**, as dependent modules have not been updated. ONLY the math objects have been moved.


z3dev:
@kaosat-dev How does this look?

kaosat-dev:
@z3dev looks good to me !
Once this is merged I think we should look at moving this whole repo into the 'main repo' as subpackages. (I have been preparing things for that)

z3dev:
There's some some rework to complete, and a few more reorganizations to execute. But yeah, let's starty here.

PR #156, opened 2019-02-11
poly3 as an array in /src/math/

pentacular:
Moves poly3 from core/geometry to math.
Changes representation of poly3 from { vertices: [a, b, c], plane: p } to x = [a, b, c]; x.plane = p.
Adds plane preserving map operation.

This allows poly3 to be used interchangeably with arrays of points, much like vec3 can be used with arrays of values.

This makes use of poly3 with interchange formats much simpler.

Justification for move from geometry to math:
  A poly3 does not support any composition operations, and seems more similar to a vec3 than to a 
  geom3 -- where vec3 is used for points, poly3 is used for face fragments.


pentacular:
Please note that this is work-in-progress.

The PR has been issued prematurely for early review -- it is not complete.

PR #157, opened 2019-02-12
Initial port of toPointsSvg algorithm.

pentacular:
An initial port of the V1 Path2D toArc line-segment approximation algorithm.


src/algorithm/arc/toPointsSvg.js
```diff
@@ -0,0 +1,113 @@
...
+ * @param {Boolean} [options.clockwise=false] - draw an arc clockwise with respect to the center point
+ * @param {Boolean} [options.large=false] - draw an arc longer than 180 degrees
+ * @param {vec2} startPoint - start of arc
+ * @param {vec2} endPoint - end of arc
+ * @returns {Array[vec2]} a point array.
+ *
+ * @example
+ * arc.toPointsSvg({ xRadius: 15, yRadius: -19.6875, xAxisRotation: 0, clockwise: false, large: false }, [0, 0], [12.5, -22.96875])
+ */
+const toPoints = ({ resolution = 32, xRadius = 0, yRadius = 0, xAxisRotation = 0, clockwise = false, large = false }, startPoint, endPoint) => {
```
z3dev: Hmm... I’m not good at names, but we can probably do better then toPoints(). I don’t like long names but this seems a lot better... generatePointsSVG()

And of course, the export and file name should match.

src/math/utils/quantizeForSpace.js
```diff
@@ -0,0 +1,6 @@
+const { spatialResolution } = require('../constants')
+
+ // Quantize values for use in spatial coordinates, and so on.
+const quantizeForSpace = (value) => (Math.round(value * spatialResolution) / spatialResolution)
```
z3dev: Wow. I looked up the definition of quantize... I don’t think this has anything to do with quants. How about roundToSpaceResolution(). Now, I know exactly what’s happening. 😃

src/algorithm/arc/toPointsSvg.js
pentacular: Yes.

I think something like 'buildSvgArc' would be a better name.
Then 'buildCircleArc' can be put in alongside it, and so on.

What do you think?

src/math/utils/quantizeForSpace.js
pentacular: Not sure what quants have to do with it. :)

The definition I've been working with is:

quan·tize
verb
1.
PHYSICS
apply quantum theory to, especially form into quanta, in particular restrict the number of possible values of (a quantity) or states of (a system) so that certain variables can assume only certain discrete magnitudes.
"light is quantized into packets of energy"
2.
ELECTRONICS
approximate (a continuously varying signal) by one whose amplitude is restricted to a prescribed set of values.
"distortion is caused when very low-level audio signals are quantized"

Let me know if that works

pentacular:
Changed name and updated V2.

PR #158, opened 2019-02-12
Syncronized core with math

z3dev:
These changes bring core and math back into syncronization again. No further movements, but this should make our continous testing sucessful going forward.

Note: Olds tests in the API directory have been disabled.


z3dev:
@kaosat-dev please do a quick review. if there are any issue then let me know.

GREEN, BABY, GREEN!

PR #166, opened 2019-04-12
add types

FishOrBear:
Add a simple type definition file.


z3dev:
@FishOrBear thanks for these changes.

It’s been a long long wait but type descriptions are now part of V2 JSCAD.

Please try V2 at www.jscad.xyz

PR #167, opened 2019-04-13
Reorganization of Geometries

z3dev:
**WORK IN PROGRESS DO NOT MERGE**

The goal of this pull request is to complete the basic geometries. Each should be complete, 100% tested, reviewed, and ready for use in other functionality.

Geometries:
- geom2
  - [x] code complete
  - [x] tests complete, 100% coverage
  - [x] reviewed
- geom3
  - [x] code complete
  - [x] tests complete, 100% coverage
  - [ ] reveiwed
- path2
  - [x] code complete
  - [x] tests complete, 100% coverage
  - [ ] reveiwed
- poly3
  - [x] code complete
  - [ ] tests complete
  - [ ] reviewed


z3dev:
@kaosat-dev please review the code for geom2, geom3, path2.

i need to add some tests to obtain 100% coverage.

are you going to make the changes to poly3 on this branch?

z3dev:
> i need to add some tests to obtain 100% coverage.

The test suites for geom2, geom3, and path2 are complete. All have 100% coverage now. :)

kaosat-dev:
I will check things out tomorrow, thanks @z3dev !
No, any changes I will make will be after this, in another branch :) (keeping things simple & manageable)

z3dev:
Some things to think about...
- Consistency... The geometries should be small and concise. Anything else, like geom2.toOutlinePaths, should go into the operations. There's some incosistency with naming as well, i.e. reverse, flip, etc.
- Parameter checking... how much should be performed? see geom2.fromPoints()
- Canonicalize... I think we should drop the canonicalize of vec2, and use full float values. It's my belief that the precision issues are due to the reuse of planes, not the point values. Planes should never be reused.

src/geometry/geom2/canonicalize.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Produces a canonicalized geometry by canonicalizing the sides.
+ * Must be called before exposing any side data.
+ * @param {geom2} geometry - the geometry to canonicalize
+ * @returns {geom2} the given geometry with transformed points
+ * @example
+ * let newgeometry = canonicalize(geometry)
+ */
+const canonicalize = (geometry) => {
```
kaosat-dev: nice and clean !

src/geometry/geom2/canonicalize.js
```diff
@@ -0,0 +1,24 @@
+const vec2 = require('../../math/vec2')
+
+/**
+ * Produces a canonicalized geometry by canonicalizing the sides.
+ * Must be called before exposing any side data.
```
kaosat-dev: I think we should be more explicit about what canonicalizing means in the first place

src/geometry/geom2/canonicalize.test.js
```diff
@@ -0,0 +1,31 @@
+const test = require('ava')
+
+const {canonicalize, fromPoints, toString} = require('./index')
+
+test('canonicalize: Updates a populated geom2 with canonalized sides', (t) => {
+  const points = [[0, 0], [1, 0], [0, 1]]
+  const expected = {
+    baseSides: [
```
kaosat-dev: what are basesides ?

src/geometry/geom2/clone.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Performs a shallow copy of the given geometry.
+ * @params {geom2} geometry - the geometry to make a shallow copy of
+ * @returns {geom2} new geometry
+ */
+const clone = (geometry) => Object.assign({}, geometry)
```
kaosat-dev: can you explain why clone is not explicit anymore ? Object.assign does not handle all copying as one would expect, so it needs to be used very carefully!

src/geometry/geom2/canonicalize.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {geom2} geometry - the geometry to canonicalize
+ * @returns {geom2} the given geometry with transformed points
+ * @example
+ * let newgeometry = canonicalize(geometry)
+ */
+const canonicalize = (geometry) => {
+  if (geometry.isCanonicalized) return geometry
+
+  // apply transforms to each side, and canonicalize
+  geometry.sides = geometry.baseSides.map((side) => {
```
kaosat-dev: why basesides ?
storing the same (or almost same) data twice in immutable data structures does not make much sense
do not forget that we have in jscad :
- obj2 = operation(obj1)
- so we do not need to keep seperate structures (very costly copies) for sides that have been transformed (transformation here being canonicalized)

src/geometry/geom2/create.js
```diff
@@ -0,0 +1,20 @@
...
+ * Create a new 2D geometry composed of unordered sides (two connected points).
+ * @param {Array} [sides] - list of sides where each side is an array of two points
+ * @returns {geom2} a new empty geometry
+ */
+const create = function (sides) {
+  if (sides === undefined) {
+    sides = [] // empty contents
+  }
+  return {
+    baseSides : sides,
```
kaosat-dev: same point as above

src/geometry/geom2/create.js
```diff
@@ -0,0 +1,20 @@
...
+    sides : [],
+    isCanonicalized : false,
+    transforms : mat4.create()
```
kaosat-dev: so transforms are combined again with geom?

src/geometry/geom2/fromPoints.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Create a new 2D geometry from the given points.
+ * The direction (rotation) of the points is not relevant,
+ * as the points can define a convex or a concave polygon.
+ * The geometry must not self intersect, i.e. the sides cannot cross.
+ * @param {Array} points - list of points in 2D space where each point is an array of two values
+ * @returns {geom2} a new geometry
+ */
+const fromPoints = function (points) {
+  if (!Array.isArray(points)) {
```
kaosat-dev: nice

src/geometry/geom2/create.test.js
```diff
@@ -0,0 +1,16 @@
+const test = require('ava')
+
+const {create} = require('./index')
+
+test('create: Creates an empty, uncanonicalized geom2', t => {
+  const expected = {baseSides: [], sides: [],
+                    isCanonicalized: false, transforms: new Float32Array([1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1]) }
```
kaosat-dev: you can use a mat4.create instead of the manual floatArray right ? or did you want it explicit ?

src/geometry/geom2/isA.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Determin if the given object is a 2D geometry.
```
kaosat-dev: Determine :)

src/geometry/geom2/isA.js
```diff
@@ -0,0 +1,17 @@
...
+ * @params {geom2} object - the object to interogate
+ * @returns {true} if the object matches a geom2 based object
+ */
+const isA = (object) => {
```
kaosat-dev: we need something more clear I think; like  'isGeom2'

src/geometry/geom2/reverse.js
```diff
@@ -0,0 +1,24 @@
+const vec2 = require('../../math/vec2')
+
+const create = require('./create')
+const toSides = require('./toSides')
+
+/**
+ * Reverses the given geometry so that the sides are flipped and in the opposite order.
+ * This swaps the left (interior) and right (exterior) edges.
```
kaosat-dev: I am not 100% sure what this means , does this change the winding order ?

src/geometry/geom2/toPoints.js
```diff
@@ -0,0 +1,24 @@
...
+ * @example
+ * let sharedpoints = toPoints(geometry)
+ */
+const toPoints = function (geometry) {
+  let sides = toSides(geometry)
+  let points = sides.map((side) => {
+    return side[0]
+  })
+  // due to the logic of fromPoints()
+  // move the first point to the last
```
kaosat-dev: is this still needed in V2 ?

src/geometry/geom2/toPoints.test.js
```diff
@@ -0,0 +1,26 @@
...
+  const geometry = create()
+  const pointarray = toPoints(geometry)
+  t.deepEqual(pointarray, [])
+})
+
+test('toPoints: creates an array of points from a populated geom2', (t) => {
+  const points = [[0, 0], [1, 0], [0, 1]]
+  const geometry = fromPoints(points)
+
+  toString(geometry)
```
kaosat-dev: is this debug ?

src/geometry/geom2/toString.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Create a string representing the contents of the given geometry.
+ * @returns {String} a representive string
+ * @example
+ * console.out(toString(geometry))
+ */
+const toString = function (geometry) {
+  if (geometry.isCanonicalized) {
+    let result = 'geom2 (' + geometry.sides.length + ' sides):\n[\n'
```
kaosat-dev: please use es6 strings instead :)

src/geometry/geom2/transform.js
```diff
@@ -0,0 +1,32 @@
+const mat4 = require('../../math/mat4')
+
+const create = require('./create')
+
+/**
+ * Transform the given geometry using the given matrix.
+ * This is a lazy transform of the sides, as this function only adjusts the transforms.
+ * See canonicalize() for the actual application of the transfroms to the sides.
```
kaosat-dev: wow ok, so with regard to my points above about baseSides: so canonicalize is needed to apply transforms ? I get the intention but that is very confusing to be honest.

kaosat-dev: canonicalize was merging 'close enough' points, I do not see how applying transforms is related ?

src/geometry/path2/concat.js
```diff
@@ -0,0 +1,30 @@
...
+ * Produces a path by concatenating the given paths.
+ * A concatenation of zero paths is an empty, open path.
+ * A concatenation of one closed path to a series of open paths produces a closed path.
+ * A concatenation of a path to a closed path is an error.
+ * @param {...path} paths - the paths to concatenate
+ * @returns {path2} new path
+ * @example
+ * let newpath = concat(fromPoints({}, [[1, 2]]), fromPoints({}, [[3, 4]]))
+ */
+const concat = (...paths) => {
```
kaosat-dev: I am scratching my head about this one, I'll get back to you on that :D

src/geometry/path2/close.js
```diff
@@ -0,0 +1,34 @@
...
+const vec2 = require('../../math/vec2')
+
+const clone = require('./clone')
+
+/**
+ * Close the given geometry.
+ * @params {geometry} the path to close
+ * @returns {path} the closed path
+ */
+const close = (geometry) => {
```
kaosat-dev: sorry If I am being dense here, but why do we need so much complexity ?
Can a closed path not be defined just as having the same start & end point ?
And would closing a path not be just appending the point if needed ?

src/geometry/path2/reverse.js
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Reverses the path so that the points are in the opposite order.
+ * This swaps the left (interior) and right (exterior) edges.
+ * Reversal of path segments with options may be non-trivial.
+ * @param {path2} path - the path to reverse.
+ * @returns {path2} the reversed path.
+ * @example
+ * reverse(path)
+ */
+const reverse = (path) => {
```
kaosat-dev: nice and simple !

src/geometry/poly3/clone.js
```diff
@@ -0,0 +1,33 @@
...
+const vec3 = require('../../math/vec3')
+
+/**
+ * Create a deep clone of the given polygon
+ *
+ * @param {vec3} [out] - receiving polygon
+ * @param {vec3} poly3 - polygon to clone
+ * @returns {vec3} clone of the polygon
+ */
+const clone = (...params) => {
```
kaosat-dev: ok

src/geometry/poly3/create.js
```diff
@@ -0,0 +1,45 @@
...
+ * Each convex polygon has a `shared` property, which is shared between all
+ *   polygons that are clones of each other or were split from the same polygon.
+ *   This can be used to define per-polygon properties (such as surface color).
+ * 
+ * The plane of the polygon is calculated from the vertex coordinates if not provided.
+ *   The plane can alternatively be passed as the third argument to avoid calculations.
+ *
+ * @constructor
+ * @param {vec3[]} vertices - list of vertices
+ * @param {shared} [shared=defaultShared] - shared property to apply
```
kaosat-dev: nice, but shared is gone right?

kaosat-dev: hmm shared was a generic data for color and more, perhaps I was wrong and we should keep it ?

src/geometry/poly3/fromPoints.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @example
+ * const points = [
+ *   [0,  0, 0],
+ *   [0, 10, 0],
+ *   [0, 10, 10]
+ * ]
+ * const polygon = fromPoints(points)
+ */
+const fromPoints = (points, planeof) => {
```
kaosat-dev: planeOf => plane

src/geometry/poly3/fromPoints.js
```diff
@@ -0,0 +1,27 @@
...
+  if (planeof) throw new Error('use fromPointAndPlane')
```
kaosat-dev: nice throwings !

src/geometry/poly3/fromPointsAndPlane.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @param {Array[]} vertices - list of vertices
+ * @param {plane} [plane] - plane of the polygon
+ */
+const fromPointsAndPlane = (vertices, plane) => {
```
kaosat-dev: nice and specific

src/geometry/poly3/isA.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Determin if the given object is a 3D polygon.
+ * @params {poly3} object - the object to interogate
+ * @returns {true} if the object matches a poly3 based object
+ */
+const isA = (object) => {
```
kaosat-dev: same remark as for others, although having a semi 'generic' operand name like this can be nice, not sure

src/geometry/poly3/measureBoundingSphere.js
```diff
@@ -0,0 +1,15 @@
+const vec3 = require('../../math/vec3')
```
kaosat-dev: I am starting to think that we could have some generic measurement functions ?
might be premature, but could we not just operate on the underlying points ?

src/geometry/geom2/canonicalize.test.js
z3dev: basePoints, baseSides, basePolygons have been removed.
i tested with the new primitives as well. :)

src/geometry/geom2/clone.js
z3dev: I depends on what we expect. Do we expect 'shallow' or 'deep' copies of the object?

math objects perform 'deep' copies.

src/geometry/geom2/canonicalize.js
z3dev: gone. please review again.

src/geometry/poly3/isA.js
z3dev: i like this idea better then having a utilitiy function that does a determination. the object itself knows best.

src/geometry/poly3/measureBoundingSphere.js
z3dev: nice idea. it would be possible now because all the geometries are point based. we will see.

src/geometry/geom2/canonicalize.js
z3dev: I'm not a fan of the term. As of NOW, the logic just 'applies' the tranforms, so I suggest we change to...

- canonicalize() ==> applyTransforms()
- isCanonicalized ==> isTransformed

src/geometry/poly3/fromPoints.js
z3dev: This is for development... I want to catch all invalid useage of this function.

src/geometry/poly3/isA.js
kaosat-dev: not sure I understand ?

src/geometry/geom2/clone.js
kaosat-dev: but should a clone not be deep by definition ? 
we never mutate things so I am not sure it matters in the end, but my point about the issues with Object.assign still stand sadly

src/geometry/geom2/canonicalize.js
kaosat-dev: I agree that the naming is better ! 
so we are clear on things
- raw geometry
- transform 
- applyTransform => sets the actual transforms, sets the 'isTransformed' flag

What happens if you apply multiple transforms ?

src/geometry/geom2/clone.js
z3dev: Good. We just need to come to the same understanding.

I will change to be ‘deep’ clones.

src/geometry/poly3/isA.js
z3dev: There were functions called isCSG and isCAG in the utilities. I think these functions should be built into the objects, i.e. isA()

src/geometry/geom2/canonicalize.js
kaosat-dev: @z3dev did you see my question above ? :)

z3dev: Sorry.

See transforms(), as that function resets the isTransformed = false, and does the matrix multiplication.

In addition, it’s really important that the contents are accessed via toPolnts, toSide, or toPolygons. Those functions will apply the transforms and returns the contents.

z3dev: @kaosat-dev Nice catch. There were some inconsistencies in the way transforms were being handled, as well as the isCanonicalized state. These were fixed. However, I would like to add test cases to make sure that the 'lazy' application REALLY works.

src/geometry/geom2/clone.js
z3dev: Done. Clone performs DEEP copies of the contents. Please review.

src/geometry/poly3/isA.js
kaosat-dev: ah yes I see your point !
hmm so we would have geom2.isA(xxx) that feels strange right ?
geom2.isOfThisType ? (meh)
isGeom2() feels more natural but you loose the reusability
then again clarity is better ?
not sure what naming is best to be honest

src/geometry/geom2/clone.js
kaosat-dev: awesome ! I'll check the updated code

kaosat-dev: ok seems good !

src/geometry/geom2/canonicalize.js
kaosat-dev: ah sorry, I had missed that !
but that makes sense, also good idea on tests

kaosat-dev: about numeric transformations and precision issues: we still need to check what happens inside of booleans and how 'very close' points are merged into a single one, but perhaps that should be another PR as it is a booleans issue

z3dev: @kaosat-dev please review geom2 only.

i added a full scenario to test lazy transforms.

also, I found something very interesting. basically, the isCanonicalized boolean is unnecessary because the transforms are always reset to the indentity matrix. the transform() function only changes the transforms matrix, while the canonicalize() function applies the transforms then resets the transforms matrix.

src/geometry/poly3/isA.js
z3dev: I kind of comes down to this...

Do we want uniquely named utility functions? isGeom2, isPath2, etc

OR

Do we want to use the namespace to call isA()? geom2.isA(), path2.isA(), etc.

You might be happier with names like geom2.isCompatible() or geom2.isEquivalentType() or geom2.isUsableAs()

src/geometry/geom2/canonicalize.js
z3dev: @kaosat-dev please review geom2 again. how do those changes look to transform() and canonicalize() ?

src/geometry/geom2/applyTransforms.js
```diff
@@ -2,14 +2,14 @@ const mat4 = require('../../math/mat4')
 const vec2 = require('../../math/vec2')
 
 /**
- * Produces a canonicalized geometry by canonicalizing the sides.
- * Must be called before exposing any side data.
- * @param {geom2} geometry - the geometry to canonicalize
- * @returns {geom2} the given geometry with transformed points
+ * Apply the transforms of the given geometry.
+ * NOTE: Must be called before exposing any side data.
```
kaosat-dev: I would change the note to something more clear like 'call this first if you need the side's data with transforms baked in'

src/geometry/geom2/toSides.js
```diff
@@ -9,7 +9,7 @@ const canonicalize = require('./canonicalize')
  * let sharedsides = toSides(geometry)
  */
 const toSides = function (geometry) {
```
kaosat-dev: ahh smart !

kaosat-dev: I really like this, makes sides data independant from data structures, well done

src/geometry/geom2/applyTransforms.js
kaosat-dev: also perhaps mention, to sides ?

z3dev:
@kaosat-dev Please review geom3 and path2, as these have been updated.
- renamed canonicalize to applyTransforms
- removed isCanonicalized as the transforms can be used to determine state
- updated all logic and tests

z3dev:
> @kaosat-dev Please review geom3 and path2, as these have been updated.

@kaosat-dev Can you review and merge this pull request?

If any changes are required then let me know. Cheers

kaosat-dev:
this all looks amazing ! thanks @z3dev  going to merge now !

PR #168, opened 2019-04-17
Reorganized text into a module

z3dev:
The goal of this pull request is to move the current text components into a module structure. The final code should be complete, 100% tested, reviewed, and ready for use in other functionality.

- text
  - [x] code complete
  - [x] tests complete, 100% coverage
  - [x] reviewed


z3dev:
@kaosat-dev please review

I believe these functions should be dropped, as this is V2. :)
- vector_char (replaced by vectorChar)
- vector_text (replaced by vectorText)

kaosat-dev:
I really like this ! I need to double check those two functions, I think there where some uses still

z3dev:
> I really like this ! I need to double check those two functions, I think there where some uses still

Maybe in the examples. But we will write new examples anyway.

z3dev:
> Maybe in the examples. But we will write new examples anyway.

Did a quick scan of the current distribution of OpenJSCAD.org. Those two functions are only found in the examples.

I think we can drop them in V2.

z3dev:
I dropped the old code and fonts.

src/text/vector_text.js
```diff
@@ -0,0 +1,18 @@
...
+* @param {Float} x - x offset
+* @param {Float} y - y offset
+* @param {String} text - ascii string
+* @returns {Array} characters segments [[[x, y], ...], ...]
+* @deprecated >= v2
+*
+* @example
+* let textSegments = vector_text(0, -20, 'OpenJSCAD')
+*/
+function vector_text (x, y, text) {
```
kaosat-dev: Ironically the naming of the functions is the only thing that I think should be changed:
- we dropped all 'snake_case' in favor of cameCase for V2 (see docs and all our changes, so the 'outward facing' api should be using that for coherence
- which brings me to a second point : why are we not just using vectorText(options, text) etc ? 
   * that is coherent for V2
   * no need to wrap anything
   * api is more rich (options object)

All the above  is also applicable for vectorChar

kaosat-dev: One thing I am not 100% sure about is whether vectorChar should be exposed or not : I do not see a difference between a string of a single caracter and a special case for just one char(vectorChar), so for simplicity I think only ### **vectorString** should be exposed

z3dev: @kaosat-dev thanks for the review.

Good. We seem to be in agreement now. The old style functions were dropped. The new style functions were retained. So, only vectorChar and vectorText remain.

You are correct about vectorChar and vectorText being equivalent. Not sure if vectorChar should be dropped as well. For now, we can keep both.

If you agree then this pull request is ready to merge.

kaosat-dev: super, thanks !
ok let's keep it for now, but I am pretty much 100% we can/should/will remove vectorChar in the end :)

kaosat-dev:
this all looks good to me @z3dev , shall I merge ?

z3dev:
please merge. :)

PR #169, opened 2019-04-26
Reorganization of Primitives

z3dev:
This pull request adds the 2D and 3D primitives as a separate module. The goals are:
- Convert to V2 math
- Standardize API across all primitives; parameters, names, etc.
- Add checks of passed parameters, and toxic combinations
- Generate 2D and 3D geometries

**Major Changes:**
- CHANGED startAngle and endAngle options to RADIAN values
- CHANGED cylinder() cylinderElliptic() roundedCylinder() to use 'height' option
- RENAMED roundedCube() to roundedCuboid()
- NEW ellipsoid() primitive
- NEW star() primitive

**Primitives:**

arc
- [X] code complete
- [X] tests complete, 93.18% coverage
- [X] review complete

circle (alias of ellipse, see ellipse.js)
- [X] code complete
- [X] tests complete

cube (alias of cuboid, see cuboid.js)
- [X] code complete
- [X] tests complete

cuboid
- [X] code complete
- [X] tests complete, 83.33% coverage
- [X] review complete

cylinder (alias of cylinderElliptic, see cylinderElliptic.js)
- [X] code complete
- [X] tests complete

cylinderElliptic
- [X] code complete
- [X] tests complete, 92.06% coverage
- [X] review complete

ellipse
- [X] code complete
- [X] tests complete, 82.76% coverage
- [X] review complete

ellipsoid
- [X] code complete
- [X] tests complete, 92.06% coverage
- [X] review complete

geodesicSphere
- [X] code complete
- [X] tests complete, 98.53% coverage
- [X] review complete

line
- [X] code complete
- [X] tests complete, 83.33% coverage
- [X] review complete

polygon
- [X] code complete
- [X] tests complete, 86.84% coverage
- [X] review complete

polyhedron
- [X] code complete
- [X] tests complete, 84.21% coverage
- [X] review complete

rectangle
- [X] code complete
- [X] tests complete, 81.82% coverage
- [X] review complete

roundedCuboid
- [X] code complete
- [X] tests complete, 98.20% coverage
- [X] review complete

roundedCylinder.js
- [X] code complete
- [X] tests complete, 94.2% coverage
- [X] review complete
Closed issue #21 as well.

roundedRectangle
- [X] code complete
- [X] tests complete, 93.75% coverage
- [X] review complete

sphere (alias of ellipsoid, see ellipsoid.js)
- [X] code complete
- [X] tests complete

square (alias of rectangle, see rectangle.js)
- [X] code complete
- [X] tests complete

star
- [X] code complete
- [X] tests complete, 91.89% coverage
- [X] review complete

torus
- [X] code complete
- [X] tests complete, 2 skipped tests
- [X] review complete
- **NOTE: DIsabled due to dependency on extrudeRotate. Should this be part of extrusions?**


z3dev:
It's time for another long discussion. I added **notes** to those primitives with issues.

It's my feeliing that primitives should be just that, and not rely upon other operations to produce geometry. For example 'torus' seems like a primitive but relies upon 'extrudeRotate'. This is a special compound shape, and may be better package in 'extrusions'.

z3dev:
General discussion: 'radius'

rectangle - radius is [x, y]
square - radius is x

Do we allow a single value for rectangle radius?
Do we allow multiple dimensions for square radius?

z3dev:
I can argue for either side of the discussion but we should be consistent.

If square is just an alias, i.e. function name, then whatever rectangle supports is what square supports. 

However, the definition of square does not allow differences in dimensions.

> A square is a regular quadrilateral, which means that it has four equal sides and four equal angles.

z3dev:
@kaosat-dev You will like this... it seems that the logic inside sphere() was actually for ellipsoid(), but had the X, Y, Z radius set to the same value. I changed things, exposing ellipsoid as a primitive.

https://en.wikipedia.org/wiki/Ellipsoid

kaosat-dev:
very cool @z3dev !
I am currently reviewing, thanks for the awesome work, will post back in a bit

src/primitives/cylinder.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {Number} [options.startRadisu=1] - radius of cylinder at the start
+ * @param {Array} [options.end=[0,1,0]] - end point of cylinder
+ * @param {Number} [options.endRadius=1] - radius of cylinder at the end
+ * @param {Number} [options.endAngle=360] - end angle of cylinder
+ * @param {Number} [options.resolution=defaultResolution3D] - number of polygons per 360 degree revolution
+ * @returns {geom3} new geometry
+ *
+ * @example
+ * let cylinder = cylinder({
+ *   start: [0, -10, 0],
```
z3dev: cylinder, cylinderElliptic, roundedCylinder use start point and end point to define the height of the cylinder. This seems counter intuitive with the other primitives which are centered, and have a Z radius. I have hit this issue many times in designs, and wonder why this was defined as such. Why not just say 'height: 10'?

also, lots of others have raised this as an issue, as the cylinder become oblique if the start and end points define a non-axis direction.

src/primitives/roundedCube.js
```diff
@@ -0,0 +1,49 @@
...
+  let innerradius = cuberadius.minus(roundradius)
+  if (innerradius.x < 0 || innerradius.y < 0 || innerradius.z < 0) {
+    throw new Error('roundradius <= radius!')
+  }
+  let res = sphere({radius: 1, resolution: resolution})
+  res = res.scale(roundradius)
+  innerradius.x > EPS && (res = res.stretchAtPlane([1, 0, 0], [0, 0, 0], 2 * innerradius.x))
+  innerradius.y > EPS && (res = res.stretchAtPlane([0, 1, 0], [0, 0, 0], 2 * innerradius.y))
+  innerradius.z > EPS && (res = res.stretchAtPlane([0, 0, 1], [0, 0, 0], 2 * innerradius.z))
+  res = res.translate([-innerradius.x + center.x, -innerradius.y + center.y, -innerradius.z + center.z])
```
kaosat-dev: same here

src/primitives/roundedRectangle.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {Number} [options.resolution=defaultResolution2D] - number of sides per 360 rotation
+ * @returns {geom2} new 2D geometry
+ *
+ * @example
+ * let myrectangle = roundedRectangle({radius: [5, 10], roundradius: 2})
+ */
+const roundedRectangle = (options) => {
+  const defaults = {
+    center: [0, 0],
+    radius: [1, 1],
```
kaosat-dev: same as for rectangle : size instead of radius

src/primitives/roundedRectangle.js
```diff
@@ -0,0 +1,57 @@
...
+
+const vec2 = require('../math/vec2')
+
+const {geom2} = require('../geometry')
+
+/** Construct a rounded rectangle.
+ * @param {Object} [options] - options for construction
+ * @param {Array} [options.center=[0,0]] - center of rounded rectangle
+ * @param {Array} [options.radius=[1,1]] - radius of rounded rectangle, width and height
+ * @param {Number} [options.roundradius=0.2] - round radius of corners
```
kaosat-dev: camelCase please

src/primitives/star.js
```diff
@@ -0,0 +1,79 @@
...
+ * @param {Number} [options.density=2] - density (Q) of star
+ * @param {Number} [options.outerRadius=1] - outer radius of vertices
+ * @param {Number} [options.innerRadius=0] - inner radius of vertices, or zero to calculate
+ * @param {Number} [options.startAngle=0] - starting angle for first vertice
+ *
+ * @example
+ * let star1 = star({vertices: 8, outerRadius: 10}) // star with 8/2 density
+ * let star2 = star({vertices: 12, outerRadius: 40, innerRadius: 20}) // star with given radius
+ */
+const star = (options) => {
```
kaosat-dev: errr , why do we need stars in 'primitives' ? that seems like something you would build using a polygon ?

src/primitives/torus.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {Integer} [options.outerResolution=defaultResolution3D] - number of segments per 360 rotation
+ * @param {Integer} [options.roti=0] - rotation angle of base circle
+ * @returns {geom3} new 3D geometry
+ *
+ * @example
+ * let torus1 = torus({
+ *   innerRadius: 10
+ * })
+ */
+function torus (options) {
```
kaosat-dev: arrrow function

src/primitives/torus.js
```diff
@@ -0,0 +1,45 @@
...
+  const defaults = {
+    innerRadius: 1,
+    innerResolution: defaultResolution2D,
```
kaosat-dev: segments ?

src/primitives/arc.js
```diff
@@ -0,0 +1,76 @@
...
+ * @returns {path} new path (not closed)
+ */
+const arc = function (options) {
+  const defaults = {
+    center: [0, 0],
+    radius: 1,
+    startangle: 0,
+    endangle: 360,
+    maketangent: false,
+    resolution: defaultResolution2D
```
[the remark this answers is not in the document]
z3dev: I renamed 'resolution' to 'segment', and really like the code read. :)

@kaosat-dev Shall I also change the constants; defaultResolution2D, defaultResolution3D?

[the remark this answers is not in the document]
z3dev: ==> it is a parameter, the default can (and should?) be specific to the shape: what do you think ?

Hmm... let me think about it. It's not a bad idea...

src/core/constants.js
```diff
@@ -39,9 +40,6 @@ const left = 3
 const right = 4
 const front = 5
 const back = 6
-// Tag factory: we can request a unique tag through CSG.getTag()
```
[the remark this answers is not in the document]
z3dev: YUP

src/primitives/roundedRectangle.js
z3dev: I'm not oppose to renaming 'radius' to 'size' but just to be sure...

radius: [x, y, z] to size: [x, y, z]
radius: N to size: N

Examples:
- circle({**size: 5,** center: [4, 3]}}
- roundedCube({**size: [5, 10, 10]**, segments: 12})

@kaosat-dev Still fine?

src/primitives/star.js
z3dev: These are difficult to create by hand, and this shape has been on my list of 'things' to add for a long long time.

FYI, I stole the code from maker.js, and converted to JSCAD.

z3dev: @danmarshall may be interested in this code as well, as there are a few improvements.

z3dev:
@kaosat-dev What to do with polygon? The API is still very SCAD-like, and very confusing. I'm not sure why the complex logic for lists-of-lists was required.

src/primitives/cylinder.js
z3dev: Posted a new discussion to the forum. Let's see if there are any opinions on this.

src/primitives/roundedRectangle.js
z3dev: What about angles... do switch to radians?

kaosat-dev:
@z3dev can you please explain a bit more about what you mean by list of lists ?

kaosat-dev:
> @kaosat-dev You will like this... it seems that the logic inside sphere() was actually for ellipsoid(), but had the X, Y, Z radius set to the same value. I changed things, exposing ellipsoid as a primitive.
> 
> https://en.wikipedia.org/wiki/Ellipsoid

Haha that is good indeed !

src/primitives/roundedRectangle.js
kaosat-dev: err wait : 
- radius/diameter is for circles, or anything based on that, because that is their mathematical definition
- size is for other type of dimentions
do not over think this :) we do not want the same parameters for all shapes, only where that makes sense !
angles in radians: yes that was the understanding we came to before right ?

src/primitives/arc.js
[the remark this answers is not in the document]
kaosat-dev: about constants: good point ! I guess it depends on whether those constants are used anywhere else ?

src/primitives/roundedRectangle.js
z3dev: Renamed radius to size for cuboid, cube, rectangle, square, roundedCube, roundedRectangle. That completes the NAME change, but the meaning of size is still 'radius'.

Do we want size to be the full size (100% width, depth, height)?
I think full 100% size makes more sense.

Back to you @kaosat-dev . What do you think?

src/primitives/arc.js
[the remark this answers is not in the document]
z3dev: Those constants (default*) should be for internal use only, so I'm starting to side with you. We should just set the default value for each primitive, which can then be more meaningful.

src/primitives/cuboid.test.js
```diff
@@ -0,0 +1,72 @@
+const test = require('ava')
+
+const geom3 = require('../geometry/geom3')
+
+const {cuboid} = require('./index')
+
+const comparePolygonsAsPoints = require('../../test/helpers/comparePolygonsAsPoints')
+
+test('cuboid (defaults)', t => {
```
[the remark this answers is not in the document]
z3dev: yup. cube takes a scalar, cuboid takes an array.

src/primitives/arc.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param {Number} options.startangle - starting angle of the arc, in degrees
+ * @param {Number} options.endangle - ending angle of the arc, in degrees
+ * @param {Number} options.resolution - number of segments to create, per 360 rotation
+ * @param {Boolean} options.maketangent - adds line segments at both ends of the arc to ensure that the gradients at the edges are tangent
+ * @returns {path} new path (not closed)
+ */
+const arc = function (options) {
+  const defaults = {
+    center: [0, 0],
+    radius: 1,
```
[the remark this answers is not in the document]
z3dev: We should just go with 'radius' for now. I think most users are working in radius already.

z3dev:
> @z3dev can you please explain a bit more about what you mean by list of lists ?

polygon() needs some help...
- it should follow polyhedron, and **ONLY** use named options; 'points' and 'paths'
- the internal logic (line 56) seems to imply that paths is an array of paths, and each path is an array of indexes. this implies that MANY polygons are combined into ONE 2D geometry. is this intended?

i think this is unsupportable as written. there are many issues with rotation, hole detection, etc.

src/primitives/cylinder.js
z3dev: As a first step, cylinder is now an alias of cylinderElliptic. That should make the final changes slightly easier.

src/primitives/polygon.js
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * let poly = polygon(roof)
+ * or
+ * let poly = polygon([roof, wall])
+ * or
+ * let poly = polygon({ points: roof })
+ * or
+ * let poly = polygon({ points: [roof, wall] })
+ * or
+ * let poly = polygon({ points: roof, paths: [0, 1, 2] })
```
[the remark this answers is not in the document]
z3dev: correct. the paths are just a series of indexes into the points. i don't see the benefit but...

z3dev:
polgyon()

> it should follow polyhedron, and **ONLY** use named options; 'points' and 'paths'
DONE

z3dev:
roundedCube() needs to be rewritten as well. There are precision errors if the resolution is adjusted to a sphere that is not planiar with X,Y,Z. For example.

`let obj = CSG.roundedCube({radius: [5,10,15], roundradius: 3, resolution: 6});`

src/primitives/cylinder.js
kaosat-dev: that is great !

src/primitives/roundedRectangle.js
kaosat-dev: size is a single 2d/3d array, I am fine with that to be honest, I think having separate width, length, height parameters might be overkill (and it would prevent passing 2d/3d arrays as inputs directly)

kaosat-dev:
@z3dev  good job on the naming of polyhedron: for now I would keep it 'as it is ' for the rest and revisit later : that function is used A LOT for all stl & other format imports, so breaking things even further can wait ? what do you think ?

about **roundedCube**: ah yes, lots of pitfalls ! (I even recall that we had notes about that one ?) 
also just thought about something : I think it should be **roundingSegments** or **roundSegments** for the rounding 'resolution' to make it clearer
About the precison errors, set of incompatible parameters: can they be caught and an error be thrown or are there realistically too many issues ?

z3dev:
> so breaking things even further can wait ? what do you think ?

Sounds like a good suggestion. Let's run with the latest. There could be a few more changes coming.

src/primitives/cylinder.js
z3dev: cylinder() cylinderElliptic() roundedCylinder() now use 'height', removing start/end points.
i like the results.

@kaosat-dev please review again.

src/primitives/roundedRectangle.js
z3dev: I still need to adjust angles to radians for...
- arc
- cylinder / cylinderElliptic
- star
- torus ('roti' option, rotation angle of base circle) rotate a circle? does that make sense?

src/primitives/arc.js
```diff
@@ -0,0 +1,76 @@
...
+
+const vec2 = require('../math/vec2')
+
+const path2 = require('../geometry/path2')
+
+/** Construct an arc.
+ * @param {Object} options - options for construction
+ * @param {Array} options.center - center of circle
+ * @param {Number} options.radius - radius of circle
+ * @param {Number} options.startangle - starting angle of the arc, in degrees
```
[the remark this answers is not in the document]
z3dev: The startAngle is now radians as well.

src/primitives/cuboid.js
```diff
@@ -0,0 +1,76 @@
+const {geom3, poly3} = require('../geometry')
+
+/** Construct an axis-aligned solid cuboid.
+ * @param {Object} [options] - options for construction
+ * @param {Array} [options.center=[0,0,0]] - center of cuboid
+ * @param {Array} [options.radius=[1,1,1]] - radius of cuboid
```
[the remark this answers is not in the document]
z3dev: This has been changed to 'size'

src/primitives/ellipsoid.js
```diff
@@ -0,0 +1,122 @@
...
+const vec3 = require('../math/vec3')
+
+const geom3 = require('../geometry/geom3')
+const poly3 = require('../geometry/poly3')
+
+/** Construct an ellipsoid.
+ * @param {Object} [options] - options for construction
+ * @param {Array} [options.center=[0,0,0]] - center of ellipsoid
+ * @param {Array} [options.radius=[1,1,1]] - radius of ellipsoid
+ * @param {Number} [options.resolution=defaultResolution3D] - number of polygons per 360 degree revolution
```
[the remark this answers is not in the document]
z3dev: This has been changed to 'segments'

src/primitives/roundedRectangle.js
kaosat-dev: thanks! awesome ! 
hmm it does seem odd like that for torus, it makes sense visually though ?

src/primitives/torus.js
```diff
@@ -0,0 +1,45 @@
...
+    outerRadius: 4,
+    outerResolution: defaultResolution3D,
+    roti: 0
```
kaosat-dev: hmm you are right, this really is not clear enough

src/primitives/star.js
kaosat-dev: hmm I understand that point, but that still contradicts the notion of specialized, 'what you need only' that we are shooting for for V2 right ?

src/primitives/cylinderElliptic.js
```diff
@@ -8,32 +8,31 @@ const {geom3, poly3} = require('../geometry')
...
  * @param {Vector2D} [options.startRadius=[1,1]] - radius of rounded start, must be two dimensional array
- * @param {Vector3} [options.end=[0,1,0]] - end point of cylinder
  * @param {Vector2D} [options.endRadius=[1,1]] - radius of rounded end, must be two dimensional array
  * @param {Number} [options.segments=12] - number of segments to create per 360 rotation
  * @returns {geom3} new geometry
  *
  * @example
  *     let cylinder = cylinderElliptic({
- *       start: [0, -10, 0],
+ *       height: 2,
```
kaosat-dev: nice

src/math/constants.js
```diff
@@ -11,6 +13,7 @@ const spatialResolution = 1e5
 const EPS = 1e-5
 
 module.exports = {
+  TWOPI,
```
kaosat-dev: could you please elaborate on this one ?

z3dev: I got tired of writing Math.Pi*2 (radians) which is the replacement for 360 (degrees)
I can revert if you don’t like it.

z3dev: Reverted.

src/primitives/torus.js
z3dev: Here's a couple of examples from V1.

```
    let obj1 = torus({ fni:4,fno:4,roti:45 });
    let obj2 = torus({ fni:4,fno:4});
```
So, 'roti' is a rotation of the circle, before creating the torus.

z3dev: If we keep this, then possibly name it... 'innerRotation'?

src/primitives/star.js
z3dev: I'd still like to keep 'star' as this is a pretty fundamental shape. The code is concise, as well as there's a matematical basis for the code.

I know that you don't want feature creep, but I would add this anway either before or after V2 is released.

z3dev:
WOW. That was one major pull request!

I'm fine with merging now.

@kaosat-dev This is now done, unless you want further changes.

kaosat-dev:
This was epic, thanks a ton @z3dev  ! WELL DONE ! merging :)

PR #170, opened 2019-04-28
Reorganization of Booleans

z3dev:
**READY FOR MERGING**

This pull request adds the boolean functions as a seperate module. The goals are:
- Convert to V2 math
- Full test coverage

This pull request adds the measurements module as well due to dependency on those functions. The goals are:
- Convert to V2 math
- Full test coverage

**Booleans:**

intersect
- [X] code complete
- [X] tests complete, 83.33% coverage (internal functions, 100% coverage)
- [X] review complete

subtract
- [X] code complete
- [X] tests complete, 83.33% coverage (internal functions, 100% coverage)
- [X] review complete

union
- [X] code complete
- [X] tests complete, 83.33% coverage (internal functions, 100% coverage)
- [X] review complete

**Measurements:**

measureArea
- [X] code complete
- [X] tests complete, 96.43% coverage
- [X] review complete

measureBounds
- [X] code complete
- [X] tests complete, 98.18% coverage
- [X] review complete

measureVolume
- [X] code complete
- [X] tests complete, 95.83% coverage
- [X] review complete


z3dev:
@kaosat-dev please review and provide feedback.

There are two modules in this pull request; measures and booleans. So, the first question is... Should these be part of the 'operations' module, or live side by side with 'math' and 'primitives'?

The two modules are structured differently. The measures have only a single file for each exposed API, and the geometry specific functions are private. The booleans have multiple files, both API and specific functions. So... Which way to package these?

z3dev:
> There are two modules in this pull request; measures and booleans. So, the first question is... Should these be part of the 'operations' module, or live side by side with 'math' and 'primitives'?

@kaosat-dev do you have any thoughts about this?

kaosat-dev:
I think it should be seperate modules ? Then again, measurements are used a lot elsewhere, but that is no issue I think :)

kaosat-dev:
Wait I missed a part of your question : what do you mean by "and the geometry specific functions are private" ?

z3dev:
> Wait I missed a part of your question : what do you mean by "and the geometry specific functions are private" ?

For example.... area() is exported in area.js. There are non-exported functions to calculate area for geom2, geom3, and path2. Those functions could also be exported, as well as broken out into specific files.

It just depends on what should be exported. Any thoughts @kaosat-dev ?

z3dev:
Next up... the 'utility' functions being used by the booleans functions.
- fromFakePolygons ; just leave here? rename to from3DWalls?
- mayOverlap ; is this useful in other modules? it doesn't really fit here, so how about moving to measures?
- retessellate ; just leave here?
- reTesselateCoplanarPolygons ; just leave here?
- to3DWalls ; just leave here?

@kaosat-dev any thoughts?

z3dev:
Logic question...

union() calls mayOverlap() in order to use the faster unionForNonIntersecting() function.

why not use the same logic for intersect() and subtract()? seems reasonable to me.

kaosat-dev:
> For example.... measureArea() is exported in measureArea.js. There are non-exported functions to calculate area for geom2, geom3, and path2. Those functions could also be exported, as well as broken out into specific files.
All things should be specific to each data/ geom type right ?

kaosat-dev:
> fromFakePolygons ; just leave here? rename to from3DWalls?

I have to take a closer look, I think there is already something named similarly to from3dWalls ?

kaosat-dev:
For all the others: yes I would leave as such and leave them there for now

kaosat-dev:
>to3DWalls
I have some doubts about the naming of that one, but for now I think it is clear enough

kaosat-dev:
> why not use the same logic for intersect() and subtract()? seems reasonable to me.

That is VERY smart thing to do !

z3dev:
> something named similarly to from3dWalls

fromFakePolygons() is actually the opposite of to3DWalls(). So, I thought we could make these into a pair of utility functions.

it seeems we were thinking the same a while back, as there is an old file called...
- ./core/geometry/geom2/fromFakeGeom3.js

z3dev:
UG!

OrthoNormalBasis is being used by reTesselateCoplanarPolygons(). Should I remove this now?
I think we can follow the suggestion in (closded) PR #160

z3dev:
> > why not use the same logic for intersect() and subtract()? seems reasonable to me.
> 
> That is VERY smart thing to do !

DONE. :)

z3dev:
@kaosat-dev please perform a code review. all code is complete, as well as initial test suites.

src/geometry/geom2/applyTransforms.js
```diff
@@ -14,8 +14,8 @@ const applyTransforms = (geometry) => {
 
   // apply transforms to each side
   geometry.sides = geometry.sides.map((side) => {
-    const p0 = vec2.canonicalize(vec2.transform(geometry.transforms, side[0]))
-    const p1 = vec2.canonicalize(vec2.transform(geometry.transforms, side[1]))
+    const p0 = vec2.transform(geometry.transforms, side[0])
```
kaosat-dev, line 17: bye bye canonicalize ? :)

src/operations/booleans/fromFakePolygons.js
```diff
@@ -0,0 +1,42 @@
+const {vec2} = require('../../math')
+
+const {geom2} = require('../../geometry')
+
+const fromFakePolygon = (polygon) => {
+  // this can happen based on union, seems to be residuals -
+  // return null and handle in caller
+  if (polygon.vertices.length < 4) {
+    return null
```
kaosat-dev, line 9: should be undefined rather than null ? (that is the case in most other parts of the updated code, although I recall a possible error in null vs undefined somewhere ?)

src/operations/booleans/fromFakePolygons.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+  var vert1Indices = []
```
kaosat-dev: var vs let vs const :)

src/operations/booleans/fromFakePolygons.js
```diff
@@ -0,0 +1,42 @@
...
+  var pts2d = polygon.vertices.filter((vertex, i) => {
+    if (vertex[2] > 0) {
+      vert1Indices.push(i)
+      return true
+    }
+    return false
+  }).map((vertex) => vec2.fromArray(vertex))
+
+  if (pts2d.length !== 2) {
+    throw new Error('Assertion failed: _fromFakePolygon: not enough points found') // TBD remove later
```
kaosat-dev: no underscore needed anymore

src/operations/booleans/fromFakePolygons.js
```diff
@@ -0,0 +1,42 @@
...
+  } else {
+    throw new Error('Assertion failed: _fromFakePolygon: unknown index ordering')
+  }
+  return pts2d
+}
+
+/** Convert the given polygons to a list of sides.
+ * The polygons must have only z coordinates +1 and -1, as constructed by to3DWalls().
+ */
+const fromFakePolygons = (polygons) => {
```
kaosat-dev, line 37: nice & simple

src/operations/booleans/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  intersect: require('./intersect'),
```
kaosat-dev, line 2: hmm the naming contradicts our existing API : difference vs subtract, intersect vs intersection, we should be coherent here

src/operations/booleans/intersect.test.js
```diff
@@ -0,0 +1,181 @@
...
+
+const comparePolygonsAsPoints = require('../../../test/helpers/comparePolygonsAsPoints')
+
+const {geom2, geom3} = require('../../geometry')
+
+const {circle, rectangle, sphere, cuboid} = require('../../primitives')
+
+const {intersect} = require('./index')
+
+//test('intersect: intersect of a path produces expected changes to points', t => {
```
kaosat-dev: we do not have this one yet ?

src/operations/booleans/mayOverlap.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+const {EPS} = require('../../core/constants')
+
+const {geom3} = require('../../geometry')
+
+const {cuboid} = require('../../primitives')
+
+const mayOverlap = require('./mayOverlap')
+
+test.only('mayOverlap: determination of overlap is perfect', t => {
```
kaosat-dev: is perfect ? perhaps something like reliable"  is a better wording?

src/operations/booleans/reTesselateCoplanarPolygons.js
```diff
@@ -0,0 +1,340 @@
...
+const OrthoNormalBasis = require('../../math/OrthoNormalBasis')
+
+const {poly3} = require('../../geometry')
+
+/**
+ * Retesselation for a set of COPLANAR polygons.
+ * @param {[poly3]} sourcepolygons - list of polygons
+ * @returns {[poly3]} new set of polygons
+ */
+const reTesselateCoplanarPolygons = (sourcepolygons) => {
```
kaosat-dev, line 14: wow !

src/operations/booleans/reTesselateCoplanarPolygons.test.js
```diff
@@ -0,0 +1,72 @@
...
+  t.is(obs.length, 1)
+  obs = reTesselateCoplanarPolygons([polyE, polyA, polyD])
+  t.is(obs.length, 1)
+
+  // combine all polygons
+  obs = reTesselateCoplanarPolygons([polyA, polyB, polyC, polyD, polyE])
+  t.is(obs.length, 1)
+
+  // now rotate everything and do again
+  let polyH = rotatePoly3([-45, -45, -45], polyA)
```
kaosat-dev, line 54: nice !!! very well done !

src/operations/booleans/subtract.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Return a new geometry representing space in the first geometry but
+ * not in all subsequent geometries.
+ * Note: None of the given geometries are modified.
+ *
+ * @param {...geometries} geometries - list of geometries
+ * @returns {geom2|geom3} a new geometry
+ *
+ * @example
+ * let myshape = subtract(cube({size: [5,5,5]}), cube({size: [5,5,5], center: [5,5,5]}))
```
kaosat-dev: cuboid ?

src/operations/booleans/subtract.js
```diff
@@ -0,0 +1,46 @@
...
+ * +----+--+    |       +----+
+ *      |   B   |
+ *      |       |
+ *      +-------+
+ */
+const subtract = (...geometries) => {
+  geometries = flatten(geometries)
+  if (geometries.length === 0) throw new Error('wrong number of arguments')
+
+  if (!areAllShapesTheSameType(geometries)) {
```
kaosat-dev, line 35: nice re-use

src/operations/booleans/to3DWalls.js
```diff
@@ -0,0 +1,24 @@
...
+  const points = [
+    vec3.fromVec2(side[0], z0),
+    vec3.fromVec2(side[1], z0),
+    vec3.fromVec2(side[1], z1),
+    vec3.fromVec2(side[0], z1)
+  ]
+  return poly3.fromPoints(points)
+}
+
+const to3DWalls = (options, geometry) => {
```
kaosat-dev, line 25: could you please add a small docstring to make it clearer what this does ? (still quite confusing when you do not know the innards :)

kaosat-dev: Also some thoughts: is a 3D 'wall' like this not just the same as the OUTLINE of a 2D shape, with Z0, Z1 set to the given values ?
meh, overthinking

src/operations/booleans/trees/Node.js
```diff
@@ -0,0 +1,139 @@
...
+// no distinction between internal and leaf nodes.
+const Node = function (parent) {
+  this.plane = null
+  this.front = null
+  this.back = null
+  this.polygontreenodes = []
+  this.parent = parent
+}
+
+Node.prototype = {
```
kaosat-dev, line 19: is this kept as is from the previous code ?
I find a bit of a bummer that we are loosing all history of the files

src/operations/booleans/trees/PolygonTreeNode.js
```diff
@@ -0,0 +1,260 @@
...
+// splitByPlane() splits a node by a plane. If the plane intersects the polygon, two new child nodes
+// are created holding the splitted polygon.
+// getPolygons() retrieves the polygon from the tree. If for PolygonTreeNode the polygon is split but
+// the two split parts (child nodes) are still intact, then the unsplit polygon is returned.
+// This ensures that we can safely split a polygon into many fragments. If the fragments are untouched,
+//  getPolygons() will return the original unsplit polygon instead of the fragments.
+// remove() removes a polygon from the tree. Once a polygon is removed, the parent polygons are invalidated
+// since they are no longer intact.
+// constructor creates the root node:
+const PolygonTreeNode = function () {
```
kaosat-dev, line 22: same as old ?

src/operations/booleans/trees/splitLineSegmentByPlane.js
```diff
@@ -0,0 +1,13 @@
+const {vec3} = require('../../../math')
+
+const splitLineSegmentByPlane = (plane, p1, p2) => {
```
kaosat-dev, line 3: does this belong here ? or maths ? (unless it is not really reuseable, then here it is fine)

src/operations/booleans/union.js
```diff
@@ -0,0 +1,44 @@
...
+const areAllShapesTheSameType = require('../../utils/areAllShapesTheSameType')
+
+const {geom2, geom3} = require('../../geometry')
+
+const unionGeom2 = require('./unionGeom2')
+const unionGeom3 = require('./unionGeom3')
+
+/**
+ * Return a new geometry representing the total space in the given geometries.
+ * NOTE: None of the given geometries are modified.
```
kaosat-dev, line 11: good

src/operations/measures/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
kaosat-dev: "measures" does not seem to be the right word, either "measuring", or "measurements" etc ?

src/operations/measures/measureBounds.js
```diff
@@ -0,0 +1,107 @@
...
+  maxpoint = [maxpoint[0], maxpoint[1], 0]
+
+  return [minpoint, maxpoint]
+}
+
+/**
+ * Measure the min and max bounds of the given (geom3) geometry.
+ * @return {[[x,y,z], [x,y,z]]} the bounds (min and max) for the geometry
+ */
+const measureBoundsOfGeom3 = (geometry) => {
```
kaosat-dev: I still have doubts about puting operations across different types in the same modules, but let's keep for now

src/operations/measures/measureVolume.js
```diff
@@ -0,0 +1,55 @@
...
+const {geom2, geom3, path2, poly3} = require('../../geometry')
+
+/*
+ * Measure the volume of the given geometry.
+ * NOTE: paths are infinitely narrow and do not have an volume
+ *
+ * @param {Path2} geometry - geometry to measure
+ * @returns {Number} volume of the geometry
+ */
+const measureVolumeOfPath2 = () => 0
```
kaosat-dev: nice !

src/operations/measures/measureVolume.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @example
+ * let volume = measureVolume(sphere())
+ */
+const measureVolume = (...geometries) => {
+  geometries = flatten(geometries)
+  if (geometries.length === 0) throw new Error('wrong number of arguments')
+
+  const results = geometries.map((geometry) => {
+    if (path2.isA(geometry)) return measureVolumeOfPath2(geometry)
```
kaosat-dev: xxx.isA is still VERY confusing to read to be honest

src/geometry/geom2/applyTransforms.js
z3dev: Agreed. Nice idea but doesn’t really solve the issues with floating point numbers. We NOW have a whole set of awesome tools to work around those issues. We just need to use them appropriately.

src/operations/booleans/reTesselateCoplanarPolygons.test.js
z3dev: :)

src/operations/booleans/trees/Node.js
z3dev: The old repository can be kept, just like Joostn's repository.

src/operations/booleans/trees/PolygonTreeNode.js
z3dev: yup

src/operations/booleans/trees/splitLineSegmentByPlane.js
z3dev: I searched the old libraries but could only find this function being used here as part of booleans.

src/operations/measures/measureVolume.js
z3dev: Yeah. I agree. I'm still hoping that one of us (or another SMART guy) will come up with a good idea.

src/operations/booleans/fromFakePolygons.js
z3dev: changed to const. done.

src/operations/booleans/fromFakePolygons.js
z3dev: done

src/operations/booleans/intersect.test.js
z3dev: Do you mean... intersect between paths?

src/operations/booleans/mayOverlap.test.js
z3dev: done

src/operations/booleans/subtract.js
z3dev: done

src/operations/booleans/to3DWalls.js
z3dev: i added documentation. done.

src/operations/booleans/fromFakePolygons.js
z3dev: I'm not sure so let's leave this for now.

src/operations/measures/index.js
z3dev: measures (noun)
- the dimensions, capacity, or amount of something ascertained by measuring 

measurements (noun)
- a figure, extent, or amount obtained by measuring

I do believe this should be a NOUN. But either is fine with me.

src/operations/booleans/index.js
z3dev: difference (NOUN, and set theory term)
subtract (VERB)

intersection (NOUN, and set theory term)
intersect (VERB)

union (NOUN, and set theory term)
?? (VERB)

By our standards, we should be using VERBS. So, 'subtract' and 'intersect' seem correct.

However, we are current using 'union' which is a NOUN. Not good. We could change to 'combine' or 'merge' or something else.

RedHatter: When Union is used to reference the mathematical operation such as in our use case, it is a verb as well as a noun. See https://en.m.wiktionary.org/wiki/union#Verb

z3dev: @RedHatter thanks for the support!

src/operations/measures/index.js
z3dev: After thinking about this for a few days, I favor the shorter 'measures'.

@kaosat-dev What do you think? Is 'measures' fine?

src/operations/booleans/index.js
z3dev: I think we have this correct. Let's leave them as VERBS.

@kaosat-dev Are you fine with this?

src/operations/booleans/fromFakePolygons.js
kaosat-dev: I this because this is in part legacy code ? In general null should be avoided for coherence imho

src/operations/booleans/fromFakePolygons.js
kaosat-dev: groovy

src/operations/measures/index.js
kaosat-dev: This is subjective I guess, but I really am not fond of measures, seems something I would find in politics. Shorter is not really essential :)

src/operations/booleans/index.js
kaosat-dev: sure, let's go with that then

src/operations/booleans/intersect.test.js
kaosat-dev: yes

src/operations/measures/index.js
z3dev: Done.
I renamed 'mesaures' to 'measurements'.

z3dev:
@kaosat-dev This is ready for merging. If you don't have any further concerns or comments then please merge.

And like always... Thanks for the excellent reviews!

kaosat-dev:
thanks for the last batch of changes, looks awesome ! Thanks for the hard work as always ! ;)

PR #171, opened 2019-04-30
Reorganization of Transforms

z3dev:
This pull request creates a small module of the transform functions. The goals are:
- Convert to V2 math
- Standardize API; function names, parameters, error checks, etc.
- Add full test suites

mirror
- [X] code complete
- [X] tests complete. 96.43% coverage
- [X] review complete

rotate
- [X] code complete
- [X] tests complete, 92.5% coverage
- [X] review complete

scale
- [X] code complete
- [X] tests complete, 86.67% coverage
- [X] review complete

transform
- [X] code complete
- [X] tests complete, 94.12% coverage
- [X] review complete

translate
- [X] code complete
- [X] tests complete, 89.29% coverage
- [X] review complete


z3dev:
The translateX, etc. functions are also defined in each file, so please review.

z3dev:
I have a BIG issue with the current rotate() function as found in V1. If you look into the code, you will find.

```
    return o.rotateX(v[0]).rotateY(v[1]).rotateZ(v[2])
```

This is the **OpenSCAD** definition of how rotate works.

If you look at the 'mutators' for rotate() then you find this definition.

```
  prot.rotate = function (rotationCenter, rotationAxis, degrees) {
    return this.transform(Matrix4x4.rotation(rotationCenter, rotationAxis, degrees))
  }
```

src/operations/transforms/mirror.js
```diff
@@ -0,0 +1,53 @@
...
+ * Mirror the given object(s) using the given options (if any)
+ * Note: The normal should be given as 90 degrees from the plane origin.
+ * @param {Object} options - options for mirror
+ * @param {Array} options.origin=[0,0,0] - the origin of the plane
+ * @param {Array} options.normal=[0,0,1] - the normal vector of the plane
+ * @param {Object|Array} objects - the objects(s) to mirror
+ * @return {Object|Array} the mirrored object(s)
+ *
+ * @example
+ * let newsphere = mirror({normal: [0,0,10]}, cube({center: [0,0,15], radius: [20, 25, 5]}))
```
kaosat-dev: nice !

src/operations/transforms/mirror.js
```diff
@@ -0,0 +1,53 @@
...
+  const defaults = {
+    origin: [0, 0, 0],
+    normal: [0, 0, 1] // Z axis
+  }
+  let {origin, normal} = Object.assign({}, defaults, options)
+
+  objects = flatten(objects)
+  if (objects.length === 0) throw new Error('wrong number of arguments')
+
+  const planeOfMirror = plane.fromNormalAndPoint(normal, origin)
```
kaosat-dev: nice use of the new stuff !

src/operations/transforms/mirror.js
```diff
@@ -0,0 +1,53 @@
...
+  const matrix = mat4.mirrorByPlane(planeOfMirror)
+
+  const results = objects.map((object) => {
+    if (path2.isA(object)) return path2.transform(matrix, object)
```
kaosat-dev: very nice use of the new functionalities ! 
However 
- it seems a bit problematic to have hard coded checks for all the data types that we have ? this means a need to change code when adding new ones (that also includes user created types): but not sure if it will really be an issue
- I am less sure about the mirroring of paths though (trying to find a good visual representation)

src/operations/transforms/mirror.js
```diff
@@ -0,0 +1,53 @@
...
+    if (geom2.isA(object)) return geom2.transform(matrix, object)
+    if (geom3.isA(object)) return geom3.transform(matrix, object)
+    return object
+  })
+  return results.length === 1 ? results[0] : results
+}
+
+const mirrorX = (...objects) => mirror({normal: [1, 0, 0]}, objects)
```
kaosat-dev: this connects back to our previous questioning: do we need mirrorX, Y , Z ? I think we could get by without them right ? (there is barely any gain , verbosity wise), perhaps we can keep them for now but flag them as candidates for removal for V2 final, no ?

src/operations/transforms/rotate.js
```diff
@@ -0,0 +1,84 @@
+const flatten = require('../utils/flatten')
+
+const mat4 = require('../../math/mat4')
+
+const {geom2, geom3, path2} = require('../../geometry')
+
+// Tait-Bryan Euler angle convention using active, intrinsic rotations around the axes in the order z-y-x
```
kaosat-dev: huh, that is a new one for me :)

kaosat-dev: ie, I do not know what it means :)

src/operations/transforms/rotate.js
```diff
@@ -0,0 +1,84 @@
...
+// angle of rotations
```
kaosat-dev: please use correct docstrings, these will not be useable for docs

src/operations/transforms/rotate.js
```diff
@@ -0,0 +1,84 @@
...
+  // create and populate rotation matrix
+  // clockwise rotation
+  //var els = [
+  //  cp*cy, sr*sp*cy - cr*sy, sr*sy + cr*sp*cy, 0,
+  //  cp*sy, cr*cy + sr*sp*sy, cr*sp*sy - sr*cy, 0,
+  //  -sp, sr*cp, cr*cp, 0,
+  //  0, 0, 0, 1
+  //]
+  // counter clockwise rotation
+  var els = [
```
kaosat-dev: no var please

src/operations/transforms/rotate.test.js
```diff
@@ -0,0 +1,149 @@
...
+      new Float32Array([ -8, -13, 18 ]), new Float32Array([ 2, -13, 18 ]) ]
+  ]
+  t.deepEqual(obs, exp)
+
+  rotated = rotateZ(180, geometry)
+  obs = geom3.toPoints(rotated)
+  t.deepEqual(obs, exp)
+})
+
+test('rotate: rotating of multiple objects produces expected changes', t => {
```
kaosat-dev: sorry but I still am not fond of the 'expected changes' formulation for tests: I do not know for example in this case what the expected should be: just apply the rotation to each object individually, or consider them as a group and use centroid of the group to rotate ?

src/operations/transforms/scale.js
```diff
@@ -0,0 +1,46 @@
...
+const {geom2, geom3, path2} = require('../../geometry')
+
+/**
+ * Scale the given object(s) using the given options (if any)
+ * @param {Array} factors - factors by which to scale the object
+ * @param {Object|Array} objects - the objects(s) to scale
+ * @return {Object|Array} the scaled object(s)
+ *
+ * @example
+ * let newsphere = scale([5, 0, 10], sphere())
```
kaosat-dev: favor const

src/operations/utils/areAllShapesTheSameType.js
```diff
@@ -0,0 +1,18 @@
+// list of supported geometries
```
kaosat-dev: to me this does not belong in operations at all but in core/common reusable code, as it is not specific to operations at all

src/operations/utils/flatten.js
```diff
@@ -0,0 +1,8 @@
+// flatten the given argument into a single flat array
```
kaosat-dev: likewise for this one, this is an array helper that belongs with other array helpers, not in operations

kaosat-dev:
about rotation: the openscad rotatation is the instinctive/ natural one: the one with a center and axis is very counter intuitive (but can have its uses), I will need to go through your new implementation to double check how that works

src/operations/transforms/rotate.js
z3dev: Here's the theorem... https://en.wikipedia.org/wiki/Euler's_rotation_theorem

Here's the paper (very good reading) that I used for the matrix logic... http://danceswithcode.net/engineeringnotes/rotations_in_3d/rotations_in_3d_part1.html

src/operations/utils/areAllShapesTheSameType.js
z3dev: i'm not even sure about the name. shouldn't this now be 'areAllGeometriesTheSame'? maybe this should be a util function as part of the geometry?

src/operations/utils/flatten.js
z3dev: i don't think there will be a core after this reorganization, but there could be some common utility functions. Should i move this to a 'utils' directory outside of operations?

fyi, there are utils in the math directory as well, but specific to the math functionality.

z3dev:
> I will need to go through your new implementation to double check how that works

Thanks. An extra pair of eyes and another brain would be good here.

I fogot to mention that the mutators also have 'rotateEulerAngles', which supports a point of rotation. I would like to add fromEulersRotation() to mat4. Is that fine?

src/operations/transforms/mirror.js
z3dev: Let's keep the mirrorX/Y/Z for now, as V1 has those functions. I have used them in designs as well.

src/operations/transforms/mirror.js
z3dev: >     * it seems a bit problematic to have hard coded checks for all the data types

I'm not a big fan either, but took the easiest route.

Definitely, we have to determine if an object is the correct type, so a function call is required to interrogate the object.

We could use some kind of pre-defined jump table to the geometry specific function, but that also means maintenance.

src/operations/transforms/rotate.test.js
z3dev: OK. I'll try to make the tests more explicit but just want to test the general functionality. In this case, I grouped a bunch of tests together.

src/operations/transforms/mirror.js
kaosat-dev: but is one of the purposes of V2 not to get rid of the clutter ? I also used it in some of my designs, but replacing it with a normal is trivial ?

src/operations/transforms/rotate.test.js
kaosat-dev: I meant only the title of the test, not the test itself :)

src/operations/utils/flatten.js
kaosat-dev: hmm good point, arrayUtils ? I honestly use the array helpers in every single design, and they are used all over in whole UI, CLI, and all other parts as well, so might as well have it in a reusable place

kaosat-dev:
hmm adding more stuff ? less is more unless really needed :D 
adding the code is one thing, but more api is another :D 
I have more time tomorrow so I can sit down and tests the rotation 'manually'

src/operations/utils/flatten.js
z3dev: Good point. Let’s move those to a more useful area for users. By the way, it also means documentation and tests need to be written.

z3dev:
> rotateEulerAngles

I think this was added/fixed a few years back by @vr050714
Again, another undocumented feature of the library but useful because it has an arbritary point of rotation. Could this be a possible replacement for OrthonormalBasis?

src/operations/utils/flatten.js
z3dev: this has been moved to src/utils/flatten.js

not perfect because the CSG library still lives outside of the BIG repository but at least a good starting point for common utility functions.

i will add some documentation and tests if this is a good compromise.

thoughts?

src/operations/transforms/rotate.test.js
z3dev: :)
Updated the titles for mirror test suites. If those look better then I will update the others test suites.

vr050714:
> > rotateEulerAngles
> 
> I think this was added/fixed a few years back by @vr050714
> Again, another undocumented feature of the library but useful because it has an arbritary point of rotation. Could this be a possible replacement for OrthonormalBasis?
> 
> I won't add the new function in this pull request but will keep the code around. :)

`OrthoNormalBasis`, if you are talking about src/math/OrthoNormalBasis.js file, might be hard to use. Hard to imagine how to get plane normal and some vector pointing to "right". Also it has an issue when this vector is collinear to normal.

`rotateEulerAngles` function is not perfect too. It must only evaluate the new basis according to given Euler's angles whereas the translation does not match to semantic of the function. Ideally, it should be like this code, for example:
```javascript
//... required imports

let rotZ1 = degToRad(90);
let rotX  = degToRad(45);
let rotZ2 = degToRad(30);

let newBasis = fromEulerAnglesZXZ(rotZ1, rotX, rotZ2);

let somePoint = [10, 20, 42];
let newPosition = translate(somePoint);

let alignedShape = transform(
                       newPosition,
                       transform(
                           newBasis,
                           originalShape
                       )
                   );
```

z3dev:
@vr050714 WOW! Thanks for the immediate feedback after... many years!

As far as OrthonormalBasis goes... we want to eliminate it. Hopefully, it won’t be difficult once we have some of the new functionality in place.

As for rotate, I think a traditional Euler angle (ZXZ) implementation is too difficult for most users to understand. Therefore, I implemented a function that uses Tait-Bryan Euler angles. Can you review the rotation() function inside of rotate.js ?

vr050714:
> @vr050714 WOW! Thanks for the immediate feedback after... many years!
> 
> As far as OrthonormalBasis goes... we want to eliminate it. Hopefully, it won’t be difficult once we have some of the new functionality in place.
> 
> As for rotate, I think a traditional Euler angle (ZXZ) implementation is too difficult for most users to understand. Therefore, I implemented a function that uses Tait-Bryan Euler angles. Can you review the rotation() function inside of rotate.js ?

Sorry, I cannot find `rotation()` function in any of 5 rotate.js files as well as a mention of either Tait-Bryan or ZYX angles. Can you point out exact file, please?

Anyway, the general purpose library must have all possible variants. And it would be awesome to have links to Wikipedia in docstrings.

src/operations/utils/areAllShapesTheSameType.js
z3dev: 'areAllGeometriesTheSame.js' has been moved to src/utils/areAllGeometriesTheSame.js

z3dev:
> Anyway, the general purpose library must have all possible variants. And it would be awesome to have links to Wikipedia in docstrings.

I found an even better reference, possibly the most complete.
https://www.wikiwand.com/en/Euler_angles#/Rotation_matrix

@vr050714 Please click on the "Files changed" tab above, then scroll down to the changes for rotate.js. Any thoughts would be welcome. :)

src/operations/transforms/rotate.js
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Rotate the given object(s) using the given options (if any)
+ * @see The paper by D. Rose, Rotations in Three-Dimensions (Tait-Bryan euler angles)
+ * @param {Object} options - options for rotate
+ * @param {Array} angles - angle of rotations about X, Y, and X axis
+ * @param {Object|Array} objects - the objects(s) to rotate
+ * @return {Object|Array} the rotated object(s)
+ *
+ * @example
+ * const newsphere = rotate({angles: [45,0,0]}, sphere())
```
vr050714: Probably, it should be `rotate([45,0,0], sphere())`

src/operations/transforms/rotate.js
```diff
@@ -0,0 +1,84 @@
...
+    sr*sy + cr*sp*cy, cr*sp*sy - sr*cy, cr*cp, 0,
+    0, 0, 0, 1
+  ]
+  return mat4.clone(els)
+}
+
+/**
+ * Rotate the given object(s) using the given options (if any)
+ * @see The paper by D. Rose, Rotations in Three-Dimensions (Tait-Bryan euler angles)
+ * @param {Object} options - options for rotate
```
vr050714: no `options` param in the code below

src/operations/transforms/rotate.js
```diff
@@ -0,0 +1,84 @@
...
+
+  // create and populate rotation matrix
+  // clockwise rotation
+  //const els = [
+  //  cp*cy, sr*sp*cy - cr*sy, sr*sy + cr*sp*cy, 0,
+  //  cp*sy, cr*cy + sr*sp*sy, cr*sp*sy - sr*cy, 0,
+  //  -sp, sr*cp, cr*cp, 0,
+  //  0, 0, 0, 1
+  //]
+  // counter clockwise rotation
```
vr050714: Matrix `rotZ = rotation(90, 0, 0)` corresponding to code below this comment corresponds to clockwise rotation of an object. It transforms vector [1, 0, 0] to [0, -1, 0] rather than [0, 1, 0]. There are two misconceptions are possible

1) What is rotated an object or a reference frame? The matrix is applied to objects.
2) What type of coordinate system is used left or right?

I think, this information should be reflected in documentation.

I did use this simple test to validate correctness of rotation
```javascript
/** Performs 4x4 matrix dot product with 4x1 vector */
function dot(mat, vect) {
  // assuming the matrixs is defined by rows
  const dotRow = (m) => [0,1,2,3].map( n => [n, 4 * m + n] )
                                 .reduce( (acc, n) => acc + mat[n[1]] * vect[n[0]], 0.0);
  return [0,1,2,3].map(dotRow);
}

function assertEqualVectors(title, actual, expected) {
  
  const tolerance = 1e-14;  // for comparison double precision values
  const isEqualVectors = [0,1,2,3].reduce((acc, n) => acc && Math.abs(actual[n] - expected[n]) < tolerance, true);

  if (isEqualVectors) {
    console.log('PASS - ' + title);
  } else {
    console.log('FAIL - ' + title);
    console.log(`    actual: [${actual}]`);
    console.log(`  expected: [${expected}]`);
  }
  console.log();
}

const unitX = [1, 0, 0, 1];
const unitY = [0, 1, 0, 1];
const unitZ = [0, 0, 1, 1];

const rotZ = rotation(90, 0,  0);
const rotX = rotation( 0, 0, 90);

assertEqualVectors('rotZ X->Y', dot(rotZ, unitX), unitY); // fail because of [0,-1,0,1]
assertEqualVectors('rotZ Z->Z', dot(rotZ, unitZ), unitZ);

assertEqualVectors('rotX Y->Z', dot(rotX, unitY), unitZ); // fail because of [0,0,-1,1]
assertEqualVectors('rotX X->X', dot(rotX, unitX), unitX);
```

z3dev: @vr050714 thanks for the comments. i feel slightly more confident now.

I think the rotation() function should be moved out, and added as mat4.fromTaitBryanAngles()
I can add tests based on the code from @vr050714

@kaosat-dev does this sound fine?

z3dev:
@vr050714 thanks again for the assistance.

I have one more request. In the V2 branch, we have a set of new math functionality. (src/math/*)
As you mentioned above, the orientation of the coordiante system is very important, especially for rotation. Therefore, I added a special test to verify that all math functionality is respecting the right-hand-rule of rotation.

Can you review src/math/rotation.test.js ?
Even now, I think some of my comments are wrong. But the rotations are correct. :)

src/operations/transforms/rotate.js
kaosat-dev: @z3dev that sound like a *very* good idea ! 
Thanks for the very good feedback  @vr050714 !

z3dev:
Pending any final comments, this pull request is ready to merge. :)

kaosat-dev:
@z3dev dang, I missed that last comment of yours: so is everything ok here from your end ? it all seems ok to me

z3dev:
Correct. Please merge.

PR #173, opened 2019-05-20
Reorganization of Hulls

z3dev:
This pull request adds the hull functions as a seperate module. The goals are:
- Convert to V2 math
- Full test coverage

hull
- [X] code complete
- [X] tests complete, 82.35% coverage (97.06% coverage for hull2DPoints)
- [X] review complete

hullChain
- [X] code complete
- [X] tests complete, 93.75% coverage
- [X] review complete


z3dev:
Is the module (directory) name correct? I think the module should be changed to 'hulls'.

@kaosat-dev Thoughts?

z3dev:
@kaosat-dev please review

src/operations/hull/hullChain.js
```diff
@@ -0,0 +1,32 @@
...
+ * Originally "Whosa whatsis" suggested "Chain Hull"
+ * @param {...geom2} geometries - list of geometries from which to create hulls
+ * @returns {geom2} new geometry
+ *
+ * @example:
+ * let newshape = hullChain(rectangle({center: [-5,-5]}), circle({center: [0,0]}), rectangle({center: [5,5]}))
+ */
+const hullChain = (options, ...geometries) => {
+  const defaults = { closed: false }
+  let { closed } = Object.assign({}, defaults, options)
```
kaosat-dev: I really wonder about the optional parameter
- it makes the api with an empty object at the start weird ?
- is the 'closed' aspect really needed ? (I am actually asking, not being ironic)

src/operations/hull/hull.test.js
```diff
@@ -0,0 +1,122 @@
+const test = require('ava')
+
+const { geom2 } = require('../../geometry')
+
+const hull = require('./hull')
+
+test('hull (single)', t => {
```
kaosat-dev: good tests !

src/operations/hull/hull.js
```diff
@@ -0,0 +1,83 @@
...
+  // assemble a new geometry from the list of points
+  return geom2.fromPoints(hullpoints)
+}
+
+/*
+ * Create a convex hull of the given geom3 geometries.
+ * @param {...geometries} geometries - list of geom3 geometries
+ * @returns {geom3} new geometry
+ */
+const hullGeom3 = (...geometries) => {
```
kaosat-dev: sigh, we really need this

src/operations/hull/hull.js
```diff
@@ -0,0 +1,83 @@
...
+ * Create a convex hull of the given geom2 geometries.
+ * @param {...geometries} geometries - list of geom2 geometries
+ * @returns {geom2} new geometry
+ */
+const hullGeom2 = (...geometries) => {
+  geometries = flatten(geometries)
+
+  // extract the unique points from the geometries
+  let uniquepoints = []
+  geometries.forEach((geometry) => {
```
kaosat-dev: since you are returning some data and not just iterating, a 'map' operation instead of forEach would be better I think ?

src/operations/hull/hull.js
```diff
@@ -0,0 +1,83 @@
...
+    let sides = geom2.toSides(geometry)
+    sides.forEach((side) => {
+      let index = uniquepoints.findIndex((unique) => vec2.equals(unique, side[0]))
+      if (index < 0) uniquepoints.push(side[0])
+    })
+  })
+
+  let hullpoints = hull2DPoints(uniquepoints)
```
kaosat-dev: for coherence I propose using hullPoints2 or heck even hullVec2s ? to match hullGeom2 ?

src/operations/hull/hull.test.js
z3dev: Thanks. I actually identified and fixed some issues with those test cases. :)

src/operations/hull/hull.js
z3dev: Hmm... I guess. But then again it would not be readable due to the nesting of interations.

src/operations/hull/hullChain.js
z3dev: Good point!

The 'closed' option just fixes the list of objects if start and end objects are different.

Question... Does the 'close' option make this function more difficutl to understand? It seems so to me.

@kaosat-dev Are you OK with removing the options?

src/operations/hull/hull.js
z3dev: it will have to wait.

src/operations/hull/hull.js
z3dev: Hmm... but the function works on 2D points. And if you search for 'hull', all documentation about convex hulls says 'set of points'. Even quickhull3D library says 'array of 3D points'.

So, I would favor a generic name so users know this function can be used with generic arrays.

kaosat-dev: I see your point, but then again we have no notion of 'point' just arrays that need to have at least two elements ?

z3dev: OK. I renamed to hullVectors2.

And expect to see hullVectors3 someday in the future. :)

z3dev:
@kaosat-dev please do a final review. and see the question above...

src/operations/hull/hullChain.js
kaosat-dev: Hmm you could just put the same shape again at the end of the list and obtain the same result right ?
I do think it makes it harder to understand WITH the option, so yes, I think removing it is ok

src/operations/hulls/hullVectors2.js
```diff
@@ -20,15 +20,15 @@ const compareIndex = (index1, index2) => {
 }
 
 // Ported from https://github.com/bkiers/GrahamScan
-const compute = (points) => {
```
kaosat-dev: @z3dev please do not hate me for this, but I think the 'points' naming was clearer, very sorry, The only thing I was not sure of is the name of the function itself, I can honestly live with the previous naming, It was me mostly thinking out lout about coherence in naming our functions, sorry !

z3dev: :) No problem. Actually, I think we should stick with a common convention anyway, which puts the 2/3 at the end. So, the correct names would be hullPoints2 / hullPoints3

src/operations/hull/hullChain.js
z3dev: Done, removed options. I updated the test suite and the documentation as well.

z3dev:
BIG NEWS! I just added hull of path2 geometries, which returns the outline (path2) of the hulled points.

z3dev:
@kaosat-dev This should be ready for merge now.

Unless... you want to add hull of geom3 geometries now.

kaosat-dev:
Looks good to me @z3dev , thanks for the hard work ! Merging

PR #177, opened 2019-07-18
V2 Performace Improvements Part 1

### LINKED ISSUE #72: Easy performance improvement - better choice of cutting plane
I recently made a bare bones port of CSG.js to C++. While implementing it, I found that simply tweaking the index that chooses which cutting plane to use while building the BSP had dramatic performance effects. Right now, the first PolygonTreeNode in the list passed to addPolygonTreeNodes is chosen. I found that picking a random index often yielded ~2x speed improvement and even picking the half way point had similar improvements. My tests were all with STL files, so I'm not sure if that skews the results in any way. Haven't tested the effect in CSG.js, but I think it's worth investigating. It makes sense as neighboring polygons are often laid out next to each other, so choosing a random polygon is more likely to provide a balanced tree.

The line of interest is here (changing 0 to just about anything else is better): https://github.com/jscad/csg.js/blob/891ad4f3f8b0eba488a7a4351d821824ab9be762/src/trees.js#L477

For those interested, my C++ port is [here](https://github.com/AllwineDesigns/stl_cmd/tree/master/src/csgjs).

z3dev:
This pull request makes performance changes across some of the highly used objects in math and geometry, as well as applying some 'known' performance improvements.

_NOTE: Also removes an reliance on old code from api or core directories._

- [X] code change complete
- [X] test coverage complete
- [X] reviews complete

Closes #72

### DIFF
--- a/src/geometry/path2/close.js
+++ b/src/geometry/path2/close.js
@@ -1,4 +1,4 @@
-const {EPS} = require('../../core/constants')
+const {EPS} = require('../../math/constants')
 
 const vec2 = require('../../math/vec2')
 
--- a/src/geometry/poly3/measureBoundingBox.js
+++ b/src/geometry/poly3/measureBoundingBox.js
@@ -4,11 +4,11 @@ const vec3 = require('../../math/vec3')
 const measureBoundingBox = (poly3) => {
   const vertices = poly3.vertices
   const numvertices = vertices.length
-  let min = numvertices === 0 ? vec3.create() : vertices[0]
-  let max = min
+  let min = numvertices === 0 ? vec3.create() : vec3.clone(vertices[0])
+  let max = vec3.clone(min)
   for (let i = 1; i < numvertices; i++) {
-    min = vec3.min(min, vertices[i])
-    max = vec3.max(max, vertices[i])
+    vec3.min(min, min, vertices[i])
+    vec3.max(max, max, vertices[i])
   }
   return [min, max]
 }
--- a/src/geometry/poly3/measureBoundingSphere.js
+++ b/src/geometry/poly3/measureBoundingSphere.js
@@ -7,7 +7,9 @@ const measureBoundingBox = require('./measureBoundingBox')
  */
 const measureBoundingSphere = (poly3) => {
   const box = measureBoundingBox(poly3)
-  const center = vec3.scale(0.5, vec3.add(box[0], box[1]))
+  const center = box[0]
+  vec3.add(center, box[0], box[1])
+  vec3.scale(center, 0.5, center)
   const radius = vec3.distance(center, box[1])
   return [center, radius]
 }
--- a/src/math/line2/fromPoints.js
+++ b/src/math/line2/fromPoints.js
@@ -10,9 +10,9 @@ const fromValues = require('./fromValues')
  * @returns {line2} a new unbounded 2D line
  */
 const fromPoints = (p1, p2) => {
-  const direction = vec2.subtract(p2, p1)
+  const normal = vec2.subtract(p2, p1)
 
-  const normal = vec2.normal(direction)
+  vec2.normal(normal, normal)
   vec2.normalize(normal, normal)
 
   const distance = vec2.dot(p1, normal)
--- a/src/math/line2/fromPoints.js
+++ b/src/math/line2/fromPoints.js
@@ -10,14 +10,14 @@ const fromValues = require('./fromValues')
  * @returns {line2} a new unbounded 2D line
  */
 const fromPoints = (p1, p2) => {
-  const normal = vec2.subtract(p2, p1)
+  const vector = vec2.subtract(p2, p1) // directional vector
 
-  vec2.normal(normal, normal)
-  vec2.normalize(normal, normal)
+  vec2.normal(vector, vector)
+  vec2.normalize(vector, vector) // normalized
 
-  const distance = vec2.dot(p1, normal)
+  const distance = vec2.dot(p1, vector)
 
-  return fromValues(normal[0], normal[1], distance)
+  return fromValues(vector[0], vector[1], distance)
 }
 
 module.exports = fromPoints
--- a/src/math/line2/fromValues.js
+++ b/src/math/line2/fromValues.js
@@ -1,3 +1,5 @@
+const create = require('./create')
+
 /**
  * Creates a new unbounded 2D line initialized with the given values.
  *
@@ -7,7 +9,7 @@
  * @returns {line2} a new unbounded 2D line
  */
 const fromValues = (x, y, w) => {
-  const out = new Float32Array(3)
+  const out = create()
   out[0] = x
   out[1] = y
   out[2] = w
--- a/src/math/line3/create.js
+++ b/src/math/line3/create.js
@@ -8,7 +8,7 @@ const fromPointAndDirection = require('./fromPointAndDirection')
  * @returns {line3} a new unbounded 3D line
  */
 const create = () => {
-  const point = vec3.fromValues(0, 0, 0)
+  const point = vec3.create() // 0, 0, 0
   const direction = vec3.fromValues(0, 0, 1)
   return fromPointAndDirection(point, direction)
 }
--- a/src/math/mat4/fromValues.js
+++ b/src/math/mat4/fromValues.js
@@ -1,3 +1,5 @@
+const create = require('./create')
+
 /**
  * Create a new mat4 with the given values
  *
@@ -20,7 +22,7 @@
  * @returns {mat4} A new mat4
  */
 const fromValues = (m00, m01, m02, m03, m10, m11, m12, m13, m20, m21, m22, m23, m30, m31, m32, m33) => {
-  const out = new Float32Array(16)
+  const out = create()
   out[0] = m00
   out[1] = m01
   out[2] = m02
--- a/src/math/plane/fromPoints.js
+++ b/src/math/plane/fromPoints.js
@@ -10,14 +10,12 @@ const fromValues = require('../vec4/fromValues')
  * @returns {Vec4} a new plane with properly typed values
  */
 const fromPoints = (a, b, c) => {
-  // let n = b.minus(a).cross(c.minus(a)).unit()
-  // FIXME optimize later
   const ba = vec3.subtract(b, a)
   const ca = vec3.subtract(c, a)
-  const cr = vec3.cross(ba, ca)
-  const normal = vec3.unit(cr) // normal part
-  const w = vec3.dot(normal, a)
-  return fromValues(normal[0], normal[1], normal[2], w)
+  vec3.cross(ba, ba, ca)
+  vec3.unit(ba, ba) // normal part
+  const w = vec3.dot(ba, a)
+  return fromValues(ba[0], ba[1], ba[2], w)
 }
 
 module.exports = fromPoints
--- a/src/math/vec2/create.js
+++ b/src/math/vec2/create.js
@@ -4,10 +4,7 @@
  * @returns {vec2} a new 2D vector
  */
 const create = () => {
-  const out = new Float32Array(2)
-  out[0] = 0
-  out[1] = 0
-  return out
+  return new Float32Array(2) // 0, 0
 }
 
 module.exports = create
--- a/src/math/vec2/fromArray.js
+++ b/src/math/vec2/fromArray.js
@@ -1,3 +1,5 @@
+const create = require('./create')
+
 /**
  * Creates a new vec2 initialized with the values in the given array
  * any value at an index > 1 is ignored !
@@ -5,7 +7,7 @@
  * @returns {vec2} a new 2D vector
  */
 const fromArray = (data) => {
-  const out = new Float32Array(2)
+  const out = create()
   out[0] = data[0]
   out[1] = data[1]
   return out
--- a/src/math/vec2/fromValues.js
+++ b/src/math/vec2/fromValues.js
@@ -1,12 +1,14 @@
+const create = require('./create')
+
 /**
- * Creates a new vec3 initialized with the given values
+ * Creates a new vec2 initialized with the given values
  *
  * @param {Number} x X component
  * @param {Number} y Y component
- * @returns {vec3} a new 2D vector
+ * @returns {vec2} a new 2D vector
  */
 const fromValues = (x, y) => {
-  const out = new Float32Array(2)
+  const out = create()
   out[0] = x
   out[1] = y
   return out
--- a/src/math/vec3/angle.js
+++ b/src/math/vec3/angle.js
@@ -1,4 +1,3 @@
-const fromValues = require('./fromValues')
 const normalize = require('./normalize')
 const dot = require('./dot')
 
@@ -9,11 +8,8 @@ const dot = require('./dot')
  * @returns {Number} The angle in radians
  */
 const angle = (a, b) => {
-  const tempA = fromValues(a[0], a[1], a[2])
-  const tempB = fromValues(b[0], b[1], b[2])
-
-  normalize(tempA, tempA)
-  normalize(tempB, tempB)
+  const tempA = normalize(a)
+  const tempB = normalize(b)
 
   const cosine = dot(tempA, tempB)
   return cosine > 1.0 ? 0 : Math.acos(cosine)
--- a/src/math/vec4/create.js
+++ b/src/math/vec4/create.js
@@ -4,12 +4,7 @@
  * @returns {vec4} a new vector
  */
 const create = () => {
-  const out = new Float32Array(4)
-  out[0] = 0
-  out[1] = 0
-  out[2] = 0
-  out[3] = 0
-  return out
+  return new Float32Array(4) // 0, 0, 0, 0
 }
 
 module.exports = create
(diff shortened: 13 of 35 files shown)


z3dev:
I will stop here as these were the first and most serious improvements for performance.

@kaosat-dev please review these changes.

src/geometry/poly3/measureBoundingBox.js
```diff
@@ -4,11 +4,11 @@ const vec3 = require('../../math/vec3')
 const measureBoundingBox = (poly3) => {
   const vertices = poly3.vertices
   const numvertices = vertices.length
-  let min = numvertices === 0 ? vec3.create() : vertices[0]
-  let max = min
+  let min = numvertices === 0 ? vec3.create() : vec3.clone(vertices[0])
```
kaosat-dev, line 7: nice

src/geometry/path2/close.js
```diff
@@ -1,4 +1,4 @@
-const {EPS} = require('../../core/constants')
+const {EPS} = require('../../math/constants')
```
kaosat-dev, line 1: are all our constants maths related ?

src/geometry/poly3/measureBoundingSphere.js
```diff
@@ -7,7 +7,9 @@ const measureBoundingBox = require('./measureBoundingBox')
  */
 const measureBoundingSphere = (poly3) => {
   const box = measureBoundingBox(poly3)
-  const center = vec3.scale(0.5, vec3.add(box[0], box[1]))
+  const center = box[0]
```
kaosat-dev, line 10: hmm why do we need to initialize, then assign to center again ?

src/math/line2/fromPoints.js
```diff
@@ -10,9 +10,9 @@ const fromValues = require('./fromValues')
  * @returns {line2} a new unbounded 2D line
  */
 const fromPoints = (p1, p2) => {
-  const direction = vec2.subtract(p2, p1)
+  const normal = vec2.subtract(p2, p1)
```
kaosat-dev: not sure what the benefit is ?

src/math/line2/fromValues.js
```diff
@@ -1,3 +1,5 @@
+const create = require('./create')
```
kaosat-dev, line 1: nice !

src/math/line3/create.js
```diff
@@ -8,7 +8,7 @@ const fromPointAndDirection = require('./fromPointAndDirection')
  * @returns {line3} a new unbounded 3D line
  */
 const create = () => {
-  const point = vec3.fromValues(0, 0, 0)
+  const point = vec3.create() // 0, 0, 0
```
kaosat-dev, line 11: nice

src/math/mat4/fromValues.js
```diff
@@ -20,7 +22,7 @@
  * @returns {mat4} A new mat4
  */
 const fromValues = (m00, m01, m02, m03, m10, m11, m12, m13, m20, m21, m22, m23, m30, m31, m32, m33) => {
-  const out = new Float32Array(16)
+  const out = create()
```
kaosat-dev, line 25: nice

src/math/plane/fromPoints.js
```diff
@@ -10,14 +10,12 @@ const fromValues = require('../vec4/fromValues')
...
 const fromPoints = (a, b, c) => {
-  // let n = b.minus(a).cross(c.minus(a)).unit()
-  // FIXME optimize later
   const ba = vec3.subtract(b, a)
   const ca = vec3.subtract(c, a)
-  const cr = vec3.cross(ba, ca)
-  const normal = vec3.unit(cr) // normal part
-  const w = vec3.dot(normal, a)
-  return fromValues(normal[0], normal[1], normal[2], w)
+  vec3.cross(ba, ba, ca)
```
kaosat-dev, line 15: this overrides the original **ba**, is that intentional?

src/math/vec2/create.js
```diff
@@ -4,10 +4,7 @@
  * @returns {vec2} a new 2D vector
  */
 const create = () => {
-  const out = new Float32Array(2)
-  out[0] = 0
-  out[1] = 0
-  return out
+  return new Float32Array(2) // 0, 0
```
kaosat-dev, line 7: good ! I double checked on MDN and TypedArrays are indeed initialized to 0

src/math/vec2/fromArray.js
```diff
@@ -1,11 +1,13 @@
...
+
 /**
  * Creates a new vec2 initialized with the values in the given array
  * any value at an index > 1 is ignored !
  * @param {Array} data array of numerical values
  * @returns {vec2} a new 2D vector
  */
 const fromArray = (data) => {
-  const out = new Float32Array(2)
+  const out = create()
```
kaosat-dev, line 10: good

src/math/vec2/fromValues.js
```diff
@@ -1,12 +1,14 @@
+const create = require('./create')
+
 /**
- * Creates a new vec3 initialized with the given values
+ * Creates a new vec2 initialized with the given values
  *
  * @param {Number} x X component
  * @param {Number} y Y component
- * @returns {vec3} a new 2D vector
+ * @returns {vec2} a new 2D vector
```
kaosat-dev, line 8: nice catches !

src/math/vec3/angle.js
```diff
@@ -9,11 +8,8 @@ const dot = require('./dot')
  * @returns {Number} The angle in radians
  */
 const angle = (a, b) => {
-  const tempA = fromValues(a[0], a[1], a[2])
-  const tempB = fromValues(b[0], b[1], b[2])
-
-  normalize(tempA, tempA)
-  normalize(tempB, tempB)
+  const tempA = normalize(a)
```
kaosat-dev, line 11: smooth !

src/math/vec4/create.js
```diff
@@ -4,12 +4,7 @@
  * @returns {vec4} a new vector
  */
 const create = () => {
-  const out = new Float32Array(4)
-  out[0] = 0
-  out[1] = 0
-  out[2] = 0
-  out[3] = 0
-  return out
+  return new Float32Array(4) // 0, 0, 0, 0
```
kaosat-dev, line 7: hehe nice :)

src/operations/booleans/intersect.test.js
```diff
@@ -170,6 +173,16 @@ test('intersect: intersect of one or more geom3 objects produces expected geomet
     [ [ 8, 8, 9 ], [ 8, 8, 8 ], [ 9, 8, 8 ], [ 9, 8, 9 ] ],
     [ [ 9, 9, 8 ], [ 9, 8, 8 ], [ 8, 8, 8 ], [ 8, 9, 8 ] ]
   ]
+*/
+  exp = [
```
kaosat-dev, line 177: hmm how come there are more points than previously ?

src/operations/booleans/trees/Node.js
```diff
@@ -89,28 +89,30 @@ Node.prototype = {
...
 
       if (polygontreenodes.length === 0) {
         current = stack.pop()
         continue
       }
-      let _this = node
       if (!node.plane) {
-        let bestplane = polygontreenodes[0].getPolygon().plane
+        let index = 0 // default
+        index = Math.floor(polygontreenodes.length / 2)
```
kaosat-dev, line 105: can you explain a bit ? not sure I understand the logic, thanks

src/operations/booleans/trees/Node.js
```diff
@@ -89,28 +89,30 @@ Node.prototype = {
...
+        //index = Math.floor(Math.random()*polygontreenodes.length)
+        let bestplane = polygontreenodes[index].getPolygon().plane
         node.plane = bestplane
       }
       let frontnodes = []
       let backnodes = []
 
       for (let i = 0, n = polygontreenodes.length; i < n; ++i) {
-        polygontreenodes[i].splitByPlane(_this.plane, _this.polygontreenodes, backnodes, frontnodes, backnodes)
+        polygontreenodes[i].splitByPlane(node.plane, node.polygontreenodes, backnodes, frontnodes, backnodes)
```
kaosat-dev, line 115: way more readable

src/operations/booleans/trees/PolygonTreeNode.js
```diff
@@ -117,7 +112,7 @@ PolygonTreeNode.prototype = {
         nodes = queue[i]
         for (j = 0, l = nodes.length; j < l; j++) { // ok to cache length
           node = nodes[j]
-          if (node.children.length) {
+          if (node.children.length > 0) {
```
kaosat-dev, line 115: carefull : that is not quite the same type of check
- if you check for 'length' you essentially check if it is an array
- if you check for length > 0 it means it is 100% always an array
Which applies in this case ?

src/operations/booleans/trees/PolygonTreeNode.js
```diff
@@ -220,6 +215,27 @@ PolygonTreeNode.prototype = {
     }
   },
 
+  clear: function () {
```
kaosat-dev, line 218: ugh, sometimes github's view is confusing, where is this one from/used ?

src/operations/booleans/trees/PolygonTreeNode.js
```diff
@@ -232,13 +248,11 @@ PolygonTreeNode.prototype = {
         node = children[j]
         result += prefix + 'PolygonTreeNode (' + node.isRootNode() + '): ' + node.children.length
         if (node.polygon) {
-          // the polygon hasn't been broken yet. We can ignore the children and return our polygon:
-          result += '\n ' + prefix + poly3.toString(node.polygon)
+          result += '\n ' + prefix + 'poly3\n'
```
kaosat-dev: es6 strings please :)

src/operations/booleans/trees/Tree.js
```diff
@@ -31,13 +31,17 @@ Tree.prototype = {
...
-    })
+    let polygontreenodes = new Array(polygons.length)
+    for (let i = 0; i < polygons.length; i++) {
+      polygontreenodes[i] = this.polygonTree.addChild(polygons[i])
+    }
     this.rootnode.addPolygonTreeNodes(polygontreenodes)
   },
 
+  clear: function () {
+    this.polygonTree.clear()
```
kaosat-dev, line 42: ah here it is :)

src/operations/booleans/trees/Tree.js
```diff
@@ -31,13 +31,17 @@ Tree.prototype = {
   },
 
   addPolygons: function (polygons) {
-    let _this = this
-    let polygontreenodes = polygons.map(function (p) {
-      return _this.polygonTree.addChild(p)
-    })
+    let polygontreenodes = new Array(polygons.length)
+    for (let i = 0; i < polygons.length; i++) {
+      polygontreenodes[i] = this.polygonTree.addChild(polygons[i])
```
kaosat-dev, line 36: any reason for the need to pre-init the polygontreenodes ? 
(speed benefit of for vs map perhaps ?)

src/operations/hulls/hullChain.test.js
```diff
@@ -30,11 +30,15 @@ test('hullChain (three, geom2)', t => {
   let obs = hullChain(geometry1, geometry2, geometry3)
   let pts = geom2.toPoints(obs)
 
-  t.is(pts.length, 10)
+  // the sides change based on the bestplane chosen in trees/Node.js
+  // t.is(pts.length, 10)
+  t.is(pts.length, 11)
```
kaosat-dev, line 35: same question as a bit above: how come the amount of points changed ?

src/operations/hulls/hullGeom2.js
```diff
@@ -16,13 +16,18 @@ const hullGeom2 = (...geometries) => {
...
   // extract the unique points from the geometries
   let uniquepoints = []
-  geometries.forEach((geometry) => {
-    let sides = geom2.toSides(geometry)
-    sides.forEach((side) => {
-      let index = uniquepoints.findIndex((unique) => vec2.equals(unique, side[0]))
-      if (index < 0) uniquepoints.push(side[0])
-    })
-  })
+  let found = []
```
kaosat-dev: ?

kaosat-dev (changes requested):
Very nice once again @z3dev ! Mostly very good, just have a few unclear point, if you could please clear them up ?

src/geometry/poly3/measureBoundingSphere.js
z3dev: Actually, center is just reusing the vec2, rather then creating (allocating) another vec2.
This function is being used in the trees and needs to be as fast as possible; PolygonTreeNode object.

src/math/line2/fromPoints.js
z3dev: The optimization consists of the elimination of one vec2 allocation. I changed the variable names, and added some comments instead.

src/operations/booleans/intersect.test.js
z3dev: Six points in each case. Just the order changes due to the choice of bestplane in Node.js.

src/operations/hulls/hullGeom2.js
z3dev: This code is actually from V1 hull(), which creates a hash from the strings. See V1 src/api/ops-transformations.js

This is about 10X faster then the nested forEach code of V2.
Figure that!

I'm not sure why there's a huge performance hit but think there may be two issues.
- object attribute are string values, and natively supported, so lookups are FAST
- closure may be occuring in the nested forEach loops, so loops are SLOW

Thoughts?

src/operations/hulls/hullChain.test.js
z3dev: The union of the hulls produces slightly different results due to the choice of bestplane.
I checked thoroughly and found that results were valid for each.

src/operations/booleans/trees/Tree.js
z3dev: correct.

src/operations/booleans/trees/Node.js
z3dev: I'm not sure 100% about this logic. But I wanted to clean up some of the poor choice of variable names, and use of _this. The code is now easier to understand.

src/operations/booleans/trees/PolygonTreeNode.js
z3dev: #2 As the change is to eliminate unnecessary push of children on line 116. The children is initially set to an empty array, so the check on line 115 is really if node.children.length is true, which is always true unless length is 0. So, I corrected the expression.

src/operations/booleans/trees/PolygonTreeNode.js
z3dev: It only gets called from Tree.clear()

FYI, I haven't determined if the clear of the tree is necessary. It's not being called at this time.

src/operations/hulls/hullGeom2.js
z3dev: If there are no solutions for the nesting of array functions then javascript kind of fails as a language. :(

src/geometry/path2/close.js
kaosat-dev: @z3dev  ?

src/geometry/poly3/measureBoundingSphere.js
kaosat-dev: hmm ok, so yes, we gain two allocation, fine by me

src/operations/booleans/trees/Node.js
kaosat-dev: I agreed with the simplification/ removal of **_this**, but this part : 
```index = Math.floor(polygontreenodes.length / 2)``` seems very different: why the use of Math.floor ? 
In the existing code it was always accessing at index 0 ?

src/operations/booleans/trees/PolygonTreeNode.js
kaosat-dev: ok , missed the initial part , that is way good :)

src/operations/hulls/hullGeom2.js
kaosat-dev: Ouch ! I am not sure if the nested for each is the reason for the slowdown: I see two more likely culprits:
- the use of findIndex for each item in the nested loop
- possible the vec2.equals aspect

I am fine with hashed but
- string based is meh
- the code readability is heavilly reduced (not your fault)

kaosat-dev:
@z3dev I added a few other comments/ replies :)

src/geometry/path2/close.js
z3dev: Correct. No more core! 😃

src/operations/hulls/hullGeom2.js
z3dev: I have some ideas and will try tomorrow.

src/operations/booleans/trees/Node.js
z3dev: See Issue #72 
This is a simple implementation, choosing the starting polygon from the middle of the list.
And, yes. There’s is a performance improvement. 😃

src/operations/hulls/hullGeom2.js
z3dev: I failed. The best solution is the nested for loops.

z3dev:
Thanks for the review. All changes have been completed.

@kaosat-dev please merge

src/operations/booleans/trees/Node.js
kaosat-dev: ah very nice !

PR #179, opened 2019-07-30
V2 - Fix of Rotate

z3dev:
It seems that rotate() was completd before we committed on RADIANS for all angles.
This pull request corrects that little error, and adjusts the test suites.

rotate:
- [X] code complete
- [X] test complete
- [X] review complete

P.S. I also fixed the GLOB parameter to ava in package.json. Way overdue.

### DIFF
--- a/src/operations/transforms/rotate.js
+++ b/src/operations/transforms/rotate.js
@@ -6,7 +6,7 @@ const { geom2, geom3, path2 } = require('../../geometry')
 
 /**
  * Rotate the given object(s) using the given options (if any)
- * @param {Array} angles - angle of rotations about X, Y, and X axis
+ * @param {[x,y,y]} angles - angle (RADIANS) of rotations about X, Y, and X axis
  * @param {Object|Array} objects - the objects(s) to rotate
  * @return {Object|Array} the rotated object(s)
  *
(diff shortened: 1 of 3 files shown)


src/operations/transforms/rotate.js
```diff
@@ -20,10 +20,9 @@ const rotate = (angles, ...objects) => {
   objects = flatten(objects)
   if (objects.length === 0) throw new Error('wrong number of arguments')
 
-  // convert to radians
-  let yaw = angles[2] * Math.PI * (1.0 / 180.0)
-  let pitch = angles[1] * Math.PI * (1.0 / 180.0)
-  let roll = angles[0] * Math.PI * (1.0 / 180.0)
+  let yaw = angles[2]
```
kaosat-dev, line 23: way simpler internals to be sure ! :)

kaosat-dev (approved):
@z3dev Sorry it took me so long to review this : it is all ok, I just kept bumping into how smart of us it is to move to radians, and I am still torn on it, but let's merge for now.
we MIGHT need to add rotateDeg in order not to turn people off ? 
I think this is the only breaking change we introduce in V2 that might cause major issues ?
Thinking also about educational aspect for kids etc
Anyway I am likely overthinking.
@z3dev shall I merge ?

z3dev:
Yeah. I know what you mean, but I’m starting to do some designs with V2, and don’t find the switch to radians that difficult. Most of the time, I’m using simple angles like 90 degrees (Math.PI/2 radians). It’s rare that I do something really difficult with angles.

z3dev:
Agreed. The switch to radians is the most significant change. But it makes the library better.

@kaosat-dev please merge.

Issue #23, opened 2017-02-26
Enhancement: getFeature: center


z3dev:
_From @dominictarr on March 15, 2016 9:40_

It would be great to be able to get the center of a openjscad shape.
That would make this useful to design, say, floating objects, i.e. boats!
You have volume, which is very useful, because you can calculate intersect a boat hull with shape representing the water, and thus calculate the displacement, if you know the center of that displacement, you can compare that to the center of the boat's mass, and know whether it is stable there. 

_Copied from original issue: joostn/OpenJsCad#81_

z3dev:
_From @dominictarr on March 15, 2016 9:57_

oh, looking closer I see I could center an object, then compare the bounds against an object that was uncentered, this would basically tell me how the center has moved.

z3dev:
_From @bebbi on March 15, 2016 22:45_

the center function would give you the geometric center. A getFeatures() function could give you the center of mass if you'd like that. See branch add_centerofmass for a quick hack. If you'd like to make this more elegant, I'm happy to take a contrib!

z3dev:
_From @dominictarr on March 16, 2016 4:45_

what is the difference between the geometric center and the center of mass? do you mean the center of mass takes into account the mass of the various parts of the model?

z3dev:
_From @bebbi on March 16, 2016 6:10_

Exactly. Imagine a boat with a long mast.
Geometric center will go right up and sit close to the middle of the mast, while center of mass stays low inside the boat. Nb the openjscad impl in that branch assumes same material ('density') used across print.

z3dev:
_From @dominictarr on March 20, 2016 22:55_

got it. to calculate the water line i'd intersect the boat model with a plane to represent the water line, water has a uniform density, so if the boat is loaded to a certain weight, it will float at that level. You don't really know the weight of the boat before building it and loading it, so it's useful to know a range... It's also important to know how far over the boat can go and still come back up, so you repeat these calculations on many different angles.

z3dev:
I updated the title to align with the work for V2, but this still remains open. We should have these additional measurements.

- measureCenter()  ; current method of finding midpoint between min and max points
- measureMidpoint() ; average of all points
- measureCenteroid ;  "centre of gravity" or the "center of mass" see http://paulbourke.net/geometry/polygonmesh/

PR #180, opened 2019-08-08
V2 : Add center parameter to cylinderElliptic / cylinder primitives

z3dev:
While converting some examples to V2, it became apparent that centering of cylinders would simplify the creation of designs. In addition, other primitives have center, so adding centering to cylinders normalizes the parameters to primitives.

cylinderElliptic:
 - [X] code complete
 - [X] tests complete
 - [X] review complete

cylinder: (alias to cylinderElliptic)
 - [X] code complete
 - [X] tests complete
 - [X] review complete


z3dev:
@kaosat-dev ... I like the idea of removing the center parameter from the primitives. Removing the center parameter will simplify the definitions for each primitive, which will help later in future plans. In addition, the center function as well as the translate functions are easily available.

Thoughts?

kaosat-dev:
@z3dev as said in hangouts, I do think that is the way forward, BUT is there any case where we need to store the center in the primitives data for any reason (I cannot think of any valid reason to do so)

z3dev:
I don't think there's any reason to store the center information as part of the primitives definition. We just need to be consistent that all primitives are 'centered' at the axis, always. This is required for symetric rotations anyway.

I think we should consider a few more measurements just to help the users. See Issue #23 
- measureCenter()  ; current method of finding midpoint between min and max points
- measureMidpoint() ; average of all points
- measureCenteroid ;  "centre of gravity" or the "center of mass" see http://paulbourke.net/geometry/polygonmesh/

kaosat-dev:
@z3dev the default positioning for cubes etc is not centered sadly (thanks to openscad) , but 100% agreed otherwise.
In addition to the measurements (I agree with all of them) we also need (I have pre-existing code too)
- align (top , bottom, on  a plane etc)
- center (the operation of centering)
- distribute

z3dev:
Cool. I’ll open new issues for all of the new functionality, one for each. And mark them for V2.

I suggest we just leave center parameters in for now, and write code and examples that use the new functionality. And, if all goes well, the center parameters are removed.

How do you want to proceed?

kaosat-dev:
So you mean adding an issue for 'removal of center' , closing this pr and remove the centers in a later pass ?

z3dev:
Correct.
Let’s remove center in the cleanup pass. Coming soon?

z3dev:
Hmm... here’s one reason for keeping center in the primitives. SVG defines circle / ellipses / etc. with a center.

`<rect x="80" y="60" width="250" height="250" rx="20"/>`

z3dev:
Autocad DXF also has the concept of center point for some ‘entities’.

`20, 30 DXF: Y and Z values of center point`

kaosat-dev:
For svg that is just a position/ translation in the end right ?
Dxf i dunno
The thing is we do not have concepts of centers for 90% of the primitives (because it does not make sense for most of them).
Here it is also only a parameter that is making a shorthand for primitive + translate.
So i think our previous discussion about not keeping them is stil valid.
(unless we get back into geometry vs transforms, haha :-)

z3dev:
agreed. now that all primitives are created 'centered on the axis'. we can translate easily.

kaosat-dev:
@z3dev cool, glad we agree on simplying :D, so what shall we do with this PR ?

z3dev:
Closing this PR based on discussions. I will save the code until V2 is done and dusted.

z3dev:
Sorry. I re-opened this pull request as I think there may be a good reason to include 'center' in primitives.

As we move forward from V2, the vision is moving to an 'expressive' set of primitives, which can have several implementations underneath. However, the expressive primitives need to be 'transformable'. In other words, the basic transforms can apply to the expressive primitives, if the position of the expressive primitive is known.

```
let geometry = circle({radius: 10, center: [-5, 10]})
geometry = geometry.translate([5, 5]) // translate can be applied to the expressive primitive
geometry = geometry.scale([0.5, 2.0]) // scale can be applied to the expressive primitive!
```

I just wanted to bring this up for discussion before putting this to sleep.

@kaosat-dev Thoughts?

z3dev:
@kaosat-dev sorry, but I think this should go into V2. From the conceptual standpoint, translation should be separate, but...

From the user point of view, centering a shape is easy, and easy to calculate as well
From the performance point, only one function to create and place the shape
From the stack point, only one data model, not two

I’d really like this to be part of V2.

kaosat-dev:
@z3dev Sorry, I should have answered here as well last time and not just on hangouts.
I have to dissagree on this, sorry: imho it is not just about adding an extra parameter into ever single primitive, this is really the same discussion as with transforms: 
- you need to store the center (which center btw ? the one for the geometry? the one for the transform, more confusion)
- what happens when you apply a transform ? what happens with the center ? 
- is it local space, model space, world space ?

>From the performance point, only one function to create and place the shape

No, under the hood, it is still multiple operations at some point or another :)

> From the stack point, only one data model, not two
not sure I understand ?

z3dev:
Closing as this was PR is part of V2.

PR #181, opened 2019-08-14
Reorganization of Connectors

z3dev:
This pull request adds the connectors as a fully exposed seperate module. The goals are:
- Convert to V2 math
- Add test suites for all functionality

Connectors:
- [X] code complete
- [X] tests complete, 100% coverage
- [X] reviewed

### DIFF
--- a/src/connectors/create.js
+++ b/src/connectors/create.js
@@ -1,18 +1,15 @@
 const vec3 = require('../math/vec3')
 
 /**
- * FIXME remove this and add to user guide
- * For example a servo motor and a servo horn:
- * Both can have a Connector called 'shaft'
- * The horn can be moved and rotated such that the two connectors match
- * and the horn is attached to the servo motor at the proper position.
- * Connectors are children of the solid, transform wise (parent => child relationship)
- * so they get the same transformations applied as the solid
- */
-
-/**
  * Create a new connector.
  * A connector allows two objects to be connected at predefined positions.
+ *
+ * For example a servo motor and a servo horn can both have a connector called 'shaft'.
+ * The horn can be moved and rotated to any position, and then the servo horn
+ * is attached to the servo motor at the proper position, such that the two connectors match.
+ * Connectors are children of the solid, transform-wise, so transformations are applied
+ * to both solid and connector(s).  (parent => child relationship)
+ *
  * @property {vec3} point - the position of the connector (relative to its parent)
  * @property {vec3} axis - the direction (unit vector) of the connector
  * @property {vec3} normal - the direction (unit vector) perpendicular to axis, that defines the "12 o'clock" orientation of the connector
--- a/src/connectors/create.js
+++ b/src/connectors/create.js
@@ -1,0 +1,30 @@
+const vec3 = require('../math/vec3')
+
+/**
+ * FIXME remove this and add to user guide
+ * For example a servo motor and a servo horn:
+ * Both can have a Connector called 'shaft'
+ * The horn can be moved and rotated such that the two connectors match
+ * and the horn is attached to the servo motor at the proper position.
+ * Connectors are children of the solid, transform wise (parent => child relationship)
+ * so they get the same transformations applied as the solid
+ */
+
+/**
+ * Create a new connector.
+ * A connector allows two objects to be connected at predefined positions.
+ * @property {vec3} point - the position of the connector (relative to its parent)
+ * @property {vec3} axis - the direction (unit vector) of the connector
+ * @property {vec3} normal - the direction (unit vector) perpendicular to axis, that defines the "12 o'clock" orientation of the connector
+ * @example
+ * let myconnector = create()
+ */
+const create = () => {
+  return {
+    point: vec3.create(),
+    axis: vec3.unit([0, 0, 1]),
+    normal: vec3.unit([1, 0, 0])
+  }
+}
+
+module.exports = create
--- a/src/connectors/extends.js
+++ b/src/connectors/extends.js
@@ -1,0 +1,14 @@
+const vec3 = require('../math/vec3')
+const fromPointAxisNormal = require('./fromPointAxisNormal')
+
+/** Creates a new Connector, with the connection point moved in the direction of the axis
+ * @param {Number} distance the distance to extend the connector to
+ * @param {Connector} connector the connector to extend
+ * @returns {Connector} a normalized connector
+ */
+const extend = (distance, connector) => {
+  const newpoint = vec3.add(connector.point, vec3.scale(distance, vec3.unit(connector.axis)))
+  return fromPointAxisNormal(newpoint, connector.axis, connector.normal)
+}
+
+module.exports = extend
--- a/src/connectors/fromPointAxisNormal.js
+++ b/src/connectors/fromPointAxisNormal.js
@@ -10,11 +10,11 @@ const create = require('./create')
  * @returns {connector} a new connector
  */
 const fromPointAxisNormal = (point, axis, normal) => {
-  let created = create()
-  created.point = vec3.fromArray(point)
-  created.axis = vec3.unit(axis)
-  created.normal = vec3.unit(normal)
-  return created
+  let connector = create()
+  connector.point = vec3.fromArray(point)
+  connector.axis = vec3.unit(axis)
+  connector.normal = vec3.unit(normal)
+  return connector
 }
 
 module.exports = fromPointAxisNormal
--- a/src/connectors/fromPointAxisNormal.js
+++ b/src/connectors/fromPointAxisNormal.js
@@ -1,0 +1,20 @@
+const vec3 = require('../math/vec3')
+
+const create = require('./create')
+
+/**
+ * Create a connector from the given point, axis and normal.
+ * @param {vec3} point - the point of the connector, relative to the parent geometry
+ * @param {vec3} axis - the axis (directional vector) of the connector
+ * @param {vec3} normal - the normal (directional vector) of the connector, perpendicular to the axis
+ * @returns {connector} a new connector
+ */
+const fromPointAxisNormal = (point, axis, normal) => {
+  let created = create()
+  created.point = vec3.fromArray(point)
+  created.axis = vec3.unit(axis)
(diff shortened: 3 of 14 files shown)


z3dev:
Sadly, this still uses OrthoNormalBasis. I'm thinking that we should kill this soon as possible. There are only two uses now.
- connectors
- reTesselateCoplanarPolygons

I think reTesselateCoplanarPolygons is overly complicated, and should be simplified. The use of to2D and to3D adds another area of imprecision to the point calculations. Maybe a task for later.

That leaves connectors using just getProjectionMatrix and getInverseProjectionMatrix. We could replace OrthoNormalBasis with the small function called toXYPlaneTransforms(). Thoughts?

src/connectors/extends.js
```diff
@@ -0,0 +1,14 @@
+const vec3 = require('../math/vec3')
+const fromPointAxisNormal = require('./fromPointAxisNormal')
+
+/** Creates a new Connector, with the connection point moved in the direction of the axis
+ * @param {Number} distance the distance to extend the connector to
+ * @param {Connector} connector the connector to extend
+ * @returns {Connector} a normalized connector
+ */
+const extend = (distance, connector) => {
```
z3dev, line 9: I don't understand the need for this function. You can do the same via transform().
Therefore, I didn't create tests and commented out of the index.

Any reason to keep this?

src/connectors/normalize.js
```diff
@@ -0,0 +1,20 @@
+const vec3 = require('../math/vec3')
+
+const fromPointAxisNormal = require('./fromPointAxisNormal')
+
+/**
+ * Normalize the given connector, calculating new axis and normal
+ * @param {Connector} connector - the connector to normalize
+ * @returns {Connector} a new connector
+ */
+const normalize = (connector) => {
```
z3dev, line 10: I don't understand the need for this function as connectors are created by only one function; fromPointAxisNormal(). And that function converts the given Axis and Normal to unit vectors.

Any reason to keep this?

src/connectors/create.js
```diff
@@ -0,0 +1,30 @@
+const vec3 = require('../math/vec3')
+
+/**
+ * FIXME remove this and add to user guide
```
kaosat-dev: I think this would also be fine as a docstring to be honest: files > wikis, as they can be shipped anywhere, included in the code etc

src/connectors/create.js
```diff
@@ -0,0 +1,30 @@
...
+ * For example a servo motor and a servo horn:
+ * Both can have a Connector called 'shaft'
+ * The horn can be moved and rotated such that the two connectors match
+ * and the horn is attached to the servo motor at the proper position.
+ * Connectors are children of the solid, transform wise (parent => child relationship)
+ * so they get the same transformations applied as the solid
+ */
+
+/**
+ * Create a new connector.
```
kaosat-dev, line 4: nice & clean

src/connectors/extends.js
kaosat-dev: I think it is not quite the same thing, (the naming is ...odd but got no better idea right now), the connector point gets moved in the *direction* of the axis, not just transformed.
I think we should leave this in for now, add it to the list of things to double check in the cleanup pass if that is ok with you ? (We will have the ability to visually compare between implementations, so that will help)

src/connectors/fromPointAxisNormal.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Create a connector from the given point, axis and normal.
+ * @param {vec3} point - the point of the connector, relative to the parent geometry
+ * @param {vec3} axis - the axis (directional vector) of the connector
+ * @param {vec3} normal - the normal (directional vector) of the connector, perpendicular to the axis
+ * @returns {connector} a new connector
+ */
+const fromPointAxisNormal = (point, axis, normal) => {
+  let created = create()
```
kaosat-dev: rather than the generic name  'created' , perhaps just 'connector' for clarity ? :)

src/connectors/fromPointAxisNormal.test.js
```diff
@@ -0,0 +1,25 @@
+const test = require('ava')
+
+const { fromPointAxisNormal, toString } = require('./index')
+
+const { compareVectors } = require('../../test/helpers/')
+
+test('connector: fromPointAxisNormal() should return a new connector with correct values', (t) => {
```
kaosat-dev, line 7: ok

src/connectors/normalize.js
kaosat-dev: I think it was/is needed in other use cases (been a while since I looked at details): can we please keep for now as well & add to cleanup pass ?

src/connectors/toString.js
```diff
@@ -0,0 +1,14 @@
...
+ * Return a string representing the given connector.
+ *
+ * @param {connector} connector - the connector of reference
+ * @returns {string} string representation
+ */
+const toString = (connector) => {
+  const point = connector.point
+  const axis = connector.axis
+  const normal = connector.normal
+  return `connector: point: [${point[0].toFixed(7)}, ${point[1].toFixed(7)}, ${point[2].toFixed(7)}],  axis: [${axis[0].toFixed(7)}, ${axis[1].toFixed(7)}, ${axis[2].toFixed(7)}], normal: [${normal[0].toFixed(7)}, ${normal[1].toFixed(7)}, ${normal[2].toFixed(7)}]`
```
kaosat-dev, line 11: yay for es6 strings , well done !
hmm just a nitpick, but could you please format it across multiple lines for readability ? (or it might be the github review page messing it up, in which case, please disregard)

src/connectors/transform.js
```diff
@@ -0,0 +1,24 @@
...
+
+const fromPointAxisNormal = require('./fromPointAxisNormal')
+
+/**
+ * Transform the give connector using the given matrix.
+ * @param {mat4} matrix - a transform matrix
+ * @param {connector} connector - the connector to transform
+ * @returns {connector} a new connector
+ */
+const transform = (matrix, connector) => {
```
kaosat-dev, line 11: good

src/connectors/transformationBetween.js
```diff
@@ -0,0 +1,64 @@
+const { mat4, plane, vec2, vec3 } = require('../math/')
+
+const OrthoNormalBasis = require('../math/OrthoNormalBasis')
+
+const transform = require('./transform')
+
+/**
+ * Get the transformation matrix that connects the given connectors.
```
kaosat-dev, line 8: nice and clear!

src/connectors/transformationBetween.js
```diff
@@ -0,0 +1,64 @@
...
+  const { mirror, normalRotation } = Object.assign({}, defaults, options)
+
+  // shift to the 0,0 origin
+  let matrix = mat4.fromTranslation(vec3.negate(from.point))
+
+  // align the axis
+  let axesplane = plane.fromPointsRandom(vec3.create(), from.axis, to.axis)
+  let axesbasis = new OrthoNormalBasis(axesplane)
+
+  let angle1 = vec2.angleRadians(axesbasis.to2D(from.axis))
```
kaosat-dev, line 32: so we have multiple things inside orthoBasis:
- to2D
- getProjectionMatrix
- getInverseProjectionMatrix

There was a proposed (but closed) PR with some alternatives to all this ?

kaosat-dev: another note: a projection matrix, is just another matrix (as in a set of operations on mat4 s) so perhaps we can do something in that direction: just adding new functions to mat4 ?

src/connectors/transformationBetween.test.js
```diff
@@ -0,0 +1,40 @@
+const test = require('ava')
+
+const { transformationBetween, create, fromPointAxisNormal } = require('./index')
+
+const { compareVectors } = require('../../test/helpers')
+
+test('connector: transformationBetween() should return correct transform matrices', (t) => {
```
kaosat-dev, line 7: wow !

src/math/OrthoNormalBasis.js
```diff
@@ -148,25 +150,24 @@ OrthoNormalBasis.Z0Plane = function () {
 }
 
 OrthoNormalBasis.prototype = {
+
   getProjectionMatrix: function () {
```
kaosat-dev, line 154: see my comments above about potentially moving these out to just functions of mat4, and bby orthoBasis !

src/math/plane/fromPointsRandom.js
```diff
@@ -11,19 +13,19 @@ const fromValues = require('../vec4/fromValues')
  * @returns {Vec4} a new plane with properly typed values
  */
 const fromPointsRandom = (a, b, c) => {
-  let v1 = vec3.subtract(b, a)
-  let v2 = vec3.subtract(c, a)
-  if (vec3.length(v1) < EPS) {
-    v1 = vec3.random(v2)
+  let ba = vec3.subtract(b, a)
```
kaosat-dev, line 16: nicer !

kaosat-dev (changes requested):
Awesome cleaning !
Only points to discuss/ change are the two functions, and the Orthobasis part (see comments in details)
Thanks :)

src/connectors/extends.js
z3dev: Ok. We can keep it. But I won’t create tests until we find a use case.

src/connectors/normalize.js
z3dev: Ok. But I won’t add tests until we find a use case.

src/connectors/fromPointAxisNormal.js
z3dev: done

src/connectors/create.js
z3dev: done

src/connectors/transformationBetween.js
z3dev: Added Issue #184 to cleanup this mess on the next pass.

z3dev:
I added a new issue to follow up on removing OrthoNormalBasis in next pass. See Issue #184 

@kaosat-dev please merge

PR #182, opened 2019-08-14
Reorganization of Expansions

### LINKED ISSUE #66: CAG expand / contract are creating degenerate CAG
The order of faces after these operations becomes wrong, and leads to unusable shapes afterwards (for extrude etc)

### LINKED ISSUE #75: Feature request: Offset for CAG
See https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Transformations#offset

### LINKED ISSUE #131: `round` parameter in `rectangular_extrude` is unsupported
The `rectangular_extrude` API supports a `round` argument but the underlying method being called (`Path2D.rectangularExtrude`) does not 

#### Code path:
- `rectangular_extrude(..., {round: false})` calls `new Path2D(...).rectangularExtrude(..., ..., ..., round)`) https://github.com/jscad/csg.js/blob/2ba12a38e020ae75404e184deed6a1aa0eccc6a0/src/api/ops-extrusions.js#L337
- `Path2D.rectangularExtrude` only supports the following three arguments `(width, height, resolution)` https://github.com/jscad/csg.js/blob/2ba12a38e020ae75404e184deed6a1aa0eccc6a0/src/core/math/Path2.js#L169

z3dev:
**WORK IN PROGRESS. DO NOT MERGE.**

This pull request adds the expansion functions as a seperate module. The goals are:
- Convert to V2 math
- Full test coverage

**Expansions:**

expand
- [X] code complete
- [X] tests complete, 89.47% coverage (87.5 / 90.0 / 84.44% coverage internally)
- [X] reveiw complete

Fixes #66 

offset
- [X] code complete
- [X] tests complete, 87.5% coverage (98.8 / 100 / 100% coverage internally)
- [X] reveiw complete

Fixes #75 
Fixes #131 
Fixes OpenJSCAD #351

Some small changes to math and geometry modulese are included.
- Added geom2.toOutlines()
- Added intersect() of two line segments to math/utils/
- Added area() under any number of ordered points to math/utils


z3dev:
Examples of offset on paths.
- black path is the original
- green path is the offset path +3 delta
- red path is the offset path -3 delta

Segments of 0 (chamfer)
offset_path_1
offset_path_2

Segments of 1 (edges meet at intersection)
offset_path_3

Segments of 32 (fillet with with rounded segments)
offset_path_4

z3dev:
Examples of offset on 2D geometries.
- black line is the original
- green line is the offset +3 delta
- red line is the offset -3 delta

Chamfer of 0 (close corners)
offset_geom2_4
offset_geom2_5
offset_geom2_6

kaosat-dev:
This is gorgeous @z3dev ! Reviewing asap !

src/geometry/geom2/toOutlines.js
```diff
@@ -0,0 +1,104 @@
...
+      nextpossiblesides.splice(nextsideindex, 1) // remove side from list
+      if (nextpossiblesides.length === 0) {
+        vertexMap.delete(nextvertex)
+      }
+      startside = nextside
+    } // inner loop
+
+    // due to the logic of fromPoints()
+    // move the first point to the last
+    if (connectedVertexPoints.length > 0) {
```
kaosat-dev, line 100: I remember this kind of logic needed from V1, is it still applicable in V2 ?

src/geometry/geom2/toOutlines.js
```diff
@@ -0,0 +1,104 @@
...
+    edges.push([getUniqueVertex(side[0]), getUniqueVertex(side[1])])
+  })
+  return edges
+}
+
+/** returns the outline(s) of the given 2D geometry as an array of points
+ * @param  {geom2} geometry
+ * @returns {Array} array of point sets, where each set is an array of points
+ */
+const toOutlines = (geometry) => {
```
kaosat-dev, line 37: I really like what this is for, lots of uses cases !

src/geometry/geom2/toOutlines.js
```diff
@@ -0,0 +1,104 @@
...
+  let edges = []
+  sides.forEach((side) => {
+    edges.push([getUniqueVertex(side[0]), getUniqueVertex(side[1])])
+  })
+  return edges
+}
+
+/** returns the outline(s) of the given 2D geometry as an array of points
+ * @param  {geom2} geometry
+ * @returns {Array} array of point sets, where each set is an array of points
```
kaosat-dev: this wording is confusing imho: (mostly the use of 'set' which is a data structure type) is this an array of arrays ? where each nested array is a 'side' ?

src/geometry/geom2/toOutlines.js
```diff
@@ -0,0 +1,104 @@
...
+        throw new Error('the given geometry is not closed. verify proper construction')
+      }
+      let nextsideindex = -1
+      if (nextpossiblesides.length === 1) {
+        nextsideindex = 0
+      } else {
+        // more than one side starting at the same vertex
+        let bestangle = undefined
+        let startangle = vec2.angleDegrees(vec2.subtract(startside[1], startside[0]))
+        for (let sideindex = 0; sideindex < nextpossiblesides.length; sideindex++) {
```
kaosat-dev, line 78: wow

src/geometry/geom2/toOutlines.test.js
```diff
@@ -0,0 +1,116 @@
...
+
+const { create, toString } = require('./index')
+
+const { comparePoints } = require('../../../test/helpers')
+
+const toOutlines = require('./toOutlines')
+
+test('geom2: toOutlines() should return no outlines for empty geom2', (t) => {
+  let shp1 = create()
+  let exp1 = []
```
kaosat-dev, line 11: this is very good ! an empty array for an empty object makes manipulating this very nice!

src/geometry/geom2/toOutlines.test.js
```diff
@@ -0,0 +1,116 @@
...
+  let ret2 = toOutlines(shp2)
+  let exp2 = [
+    [[5, -5], [6, -5], [6, -4]],
+    [[-10, -10], [10, -10], [10, 10]],
+    [[8, 6], [8, -8], [-6, -8]]
+  ]
+  t.true(comparePoints(exp2, ret2))
+})
+
+test('geom2: toOutlines() should return outlines for edges that touch in geom2', (t) => {
```
kaosat-dev, line 72: this is very hard to visualize: what are 'touching' edges ?

src/math/utils/area.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * Calculate the area under the given points
+ */
+const area = (points) => {
```
kaosat-dev, line 4: nice

src/math/utils/area.js
```diff
@@ -0,0 +1,14 @@
...
+  let area = 0
+  for (let i = 0; i < points.length; i++) {
+    let j = (i + 1) % points.length
+    area += points[i][0] * points[j][1]
+    area -= points[j][0] * points[i][1]
+  }
+  return (area / 2.0)
+}
+
+module.exports = area
```
kaosat-dev, line 14: hmm thinking a bit more about this: this seems specifically for Vec2s (2d data) so I think it belongs there ?

src/math/utils/intersect.js
```diff
@@ -0,0 +1,33 @@
+// Calculate the intersect point of the two line segments (p1-p2 and p3-p4), end points included.
+// @see http://paulbourke.net/geometry/pointlineplane/
+// Note: If the line segments do NOT intersect then undefined is returned
+// Return the intersection point of two line segments
+const intersect = (p1, p2, p3, p4) => {
```
kaosat-dev, line 11: usefull ! nice !

src/math/utils/intersect.js
```diff
@@ -0,0 +1,33 @@
+// Calculate the intersect point of the two line segments (p1-p2 and p3-p4), end points included.
```
kaosat-dev: please switch to normal docstrings

src/math/utils/intersect.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  // Return the x and y coordinates of the intersection
+  let x = p1[0] + ua * (p2[0] - p1[0])
+  let y = p1[1] + ua * (p2[1] - p1[1])
+
+  return [x, y]
+}
+
+module.exports = intersect
```
kaosat-dev, line 39: hmm I am scratching my head over where this would fit best: it is also specific for 2D data, not sure what the right place would be ?

src/operations/expansions/expand.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {Object} options - options for expand
+ * @param {Number} options.delta=1 - delta (+/-) of expansion
+ * @param {Integer} [options.segments] - number of segments when creating rounded corners, or zero for chamfer
+ * @param {Object|Array} objects - the objects(s) to expand
+ * @return {Object|Array} the expanded object(s)
+ *
+ * @example
+ * let newsphere = expand({delta: 2}, cube({center: [0,0,15], size: [20, 25, 5]}))
+ */
+const expand = (options, ...objects) => {
```
kaosat-dev, line 21: clean & awesome !

src/operations/expansions/expandGeom2.js
```diff
@@ -0,0 +1,20 @@
+const { geom2 } = require('../../geometry')
+
+const offsetGeom2 = require('./offsetGeom2')
+
+/**
+ * Expand the given geometry (geom2) using the given options (if any).
+ * @param {Object} options - options for expand
+ * @param {Number} [options.delta=1] - delta (+/-) of expansion
+ * @param {Integer} [options.segments=0] - number of segments when creating rounded corners, or zero for chamfer
```
kaosat-dev: hmm I both like this (using segments === 0 for chamfer) , if we had a non polygon based system, would you say that passing Infinity (an actual js thing) to this value would generate something with maximum resolution/ no polygons?

src/operations/expansions/expandPath2.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Expand the given geometry (path2) using the given options (if any).
+ * @param {Object} options - options for expand
+ * @param {Number} [options.delta=1] - delta (+) of expansion
+ * @param {Integer} [options.segments=0] - number of segments when creating rounded corners, or zero for chamfer
+ * @param {path2} geometry - the geometry to expand
+ * @returns {geom2} expanded geometry
+ */
+const expandPath2 = (options, geometry) => {
```
kaosat-dev, line 18: very awesome ! (despite my brain still breaking when thinking about what an expansion of a non closed path means ?)

src/operations/expansions/expandShell.js
```diff
@@ -0,0 +1,215 @@
...
+const { geom3, poly3 } = require('../../geometry')
+
+const { sphere } = require('../../primitives')
+
+const retessellate = require('../booleans/retessellate')
+const unionGeom3Sub = require('../booleans/unionGeom3Sub')
+
+const extrudePolygon = require('./extrudePolygon')
+
+const fnNumberSort = (a, b) => {
```
kaosat-dev: don't we have that in the utils ?

src/operations/expansions/extrudePolygon.js
```diff
@@ -0,0 +1,32 @@
+const { mat4, vec3 } = require('../../math')
+
+const { geom3, poly3 } = require('../../geometry')
+
+// Extrude a polygon in the direction of the offsetvector.
+// Returns (geom3) a new geometry
+const extrudePolygon = (offsetvector, polygon1) => {
```
kaosat-dev, line 7: This is unrelated to extrusions I am guessing ? Should it not be 'offsetPolygon' ? (just asking, drawing a blank)

src/operations/expansions/inside.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/* Determine if the given point is inside the polygon.
+ *
+ * @see http://paulbourke.net/geometry/polygonmesh/
+ * @param {Array} point - an array with X and Y values
+ * @param {Array} polygon - a list of points, where each point is an array with X and Y values
+ * @return {Integer} 1 if the point is inside, 0 if outside
+ */
+const isPointInside = (point, polygon) => {
```
kaosat-dev, line 21: these two are great, but I think they belong in a more generic place, for reuse (over the years I had a need for these a lot in designs

test/helpers/index.js
```diff
@@ -1,5 +1,7 @@
 module.exports = {
+  comparePoints: require('./comparePoints'),
   comparePolygons: require('./comparePolygons'),
+  comparePolygonsAsPoints: require('./comparePolygonsAsPoints'),
```
kaosat-dev, line 4: I do not see this added as part of this PR ? or am I missing something

src/operations/expansions/offset.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Create offset geometry(s) from the given object(s) using the given options (if any).
+ * @param {Object} options - options for offset
+ * @param {Float} [options.delta=1] - delta of offset (+ to exterior, - from interior)
+ * @param {Integer} [options.segments=0] - segments of rounded ends, or zero for chamfer
+ * @param {Object|Array} objects - object(s) to offset
+ * @return {Object|Array} the offset geometry(s)
+ */
+const offset = (options, ...objects) => {
```
kaosat-dev, line 17: ok I confess I am really confused: even going back & forth between files and your screenshots, I do not understand expand vs offset ?

kaosat-dev (changes requested):
Top notch work and very eager to try this out as a user @z3dev !
There are a few points I am unclear on however if you do not mind ?

src/operations/expansions/offset.js
z3dev: Offset 
- path2 ; create offset of path with chamfers
- geom2 ; convert to outlines, offset each outline, combine points into geom2

Expand
- path2 ; create offsets on both sides, create ends, and convert to geom2
- geom2 ; same as offset
- geom3 ; old logic and very very slow

We could do more, but I stopped at this point.

src/operations/expansions/inside.js
z3dev: Sure. This is pretty generic. The polygon is defined as an ordered list of points

@kaosat-dev  So how about moving to utils?

src/operations/expansions/extrudePolygon.js
z3dev: I’ll look again. It may be in the wrong place.

src/geometry/geom2/toOutlines.js
z3dev: I’ll review again.

src/operations/expansions/expandShell.js
z3dev: It will be when the PR for performance is merged. :)

src/geometry/geom2/toOutlines.js
z3dev: Yes. This is still required as geom2.fromPoints() creates the first side from points[N-1] and points[0].

src/operations/expansions/extrudePolygon.js
z3dev: extrudePolygon() is being used by expandShell() for geom3 expansion. it's included here as I found no other use of this function in the library.

test/helpers/index.js
z3dev: Just updating the index. The helpler function has been there for a long time.

src/geometry/geom2/toOutlines.js
z3dev: Updated in-line documentation.

src/geometry/geom2/toOutlines.test.js
z3dev: Like a figure 8... where there's a point of intersection between two sides.

src/math/utils/intersect.js
z3dev: maybe the name should change? intersectLineSegments?

src/operations/expansions/expandGeom2.js
z3dev: Yes. I think you are correct.
Maybe we would use 'fill' instead of segments? Where 'fill' means:
- 0 : close corners with a chamfer
- 1 : close corners with an edge
- N : close corners with a fillet (of course, we use count of segments now)

src/operations/expansions/expandPath2.js
z3dev: I suggest we play with this once integrated into the viewer. It will be far easier to perceive, and changes may be needed.

src/operations/expansions/extrudePolygon.js
z3dev: @kaosat-dev should I move extrudePolygon to extrusions?

src/operations/expansions/expandShell.js
z3dev: Done

src/operations/expansions/extrudePolygon.js
kaosat-dev: not sure to be honest : is it actual extrusion ? Ie like taking a flat shape and turning it 3D ?

src/operations/expansions/inside.js
kaosat-dev: hmm I agree about generic, but 'utils' seems too generic: this will only work for 2D points, so I would
- put it under geom2 since we do not have 2d polygons as data structures? 
- rename the file to a more clear arePointsInsidePoly2.js

src/operations/expansions/offset.js
kaosat-dev: sorry for the slow reply on this one: if expand does the same as offset for geom2 it means to me that it is the same operand, so this seems very confusing as it is to be honest
it seems that the main issue here is expand/ offset for path2 since that is the only one that has different results.
I also have a hard time understanding (sorry!) why chamfer is a part of offset and not a different operation ?
I honestly like the results a LOT, but I really think more thought around the API is really worth it for this :)

z3dev: ‘Chamfer’ is my term for how to fill the corners, but I think this term makes sense. And we could use the term for rounded rectangles and rounded cubes as well. ‘Segments’ isn’t really the correct term.

https://en.m.wikipedia.org/wiki/Chamfer

The offset() function creates parallel line segments, but that leaves gaps on the corners. The ‘chamfer’ tells the routine how to fill those gaps.

Although, a chamfer() function may be possible I’m not sure it makes sense because the offset() function creates the gaps, and tracks the gaps. Its far more efficient to fill those gaps as part of offset().

z3dev:
250E4424-9176-49C6-8BB0-D4B0EE10CC1B

(This is actually from a wood working website.)

src/operations/expansions/extrudePolygon.js
z3dev: Correct. It uses the polygon as the top and the bottom, and builds sides. The result is a new 3D geometry.

src/operations/expansions/inside.js
z3dev: @kaosat-dev how about creating a poly2 data structure? The only uses now are:
- measureArea (now called area() is math utils)
- arePointsInside

It seems kind of a waste but really these functions would better align as part of this data structure. And who knows, later enhancements might require poly2.

src/operations/expansions/extrudePolygon.js
z3dev: Let’s leave it here for now.

src/operations/expansions/offset.js
z3dev: RESET

z3dev:
For CAG V1, here's what expand() does to a shape with a hole (10x10 square with 5x5 hole). Notice that the hole is getting smaller, not larger.
```
    let obj1 = CAG.rectangle({radius: [10,10]});
    obj1 = obj1.subtract(CAG.rectangle({radius: [5,5]}));
    
    let obj2 = obj1.expand(1.0, 8);
```
V1expand

z3dev:
For V2, I think the same functionality should be offset. (See expand below)
```
  let obj1 = primitives.square({size: 10})
  obj1 = booleans.subtract(obj1, primitives.rectangle({size: [5, 5], center: [0,0]}))
  obj1.color = [1,1,0,1] // YELLOW

  let obj2 = expansions.offset({delta: 1}, obj1)
  obj2.color = [0,1,0,1] // GREEN
```
V2offset

z3dev:
For V2, the expand makes the whole shape expand (enlarge), including the holes. Note, this is very simular to scale however scale cannot chamfer corners.
```
  let obj1 = primitives.square({size: 10})
  obj1 = booleans.subtract(obj1, primitives.rectangle({size: [5, 5], center: [0,0]}))
  obj1.color = [1,1,0,1] // YELLOW

  let obj2 = expansions.expand({delta: 1}, obj1)
  obj2.color = [0,1,0,1] // GREEN
```
V2expand

src/operations/expansions/expandPath2.js
z3dev: @kaosat-dev do you remember this little piece of wonder?
```
    // Expand the path to a CAG
    // This traces the path with a circle with radius pathradius
  expandToCAG: function (pathradius, resolution)
```
So, expanding a path is not new. BUT, a whole lot more sensical.

src/operations/expansions/offset.js
kaosat-dev: ?? reset

z3dev: I made some changes so let’s start again, with a new discussion.

src/operations/expansions/inside.js
kaosat-dev: ok, agreed, not perfect, but coherent, and we can always change later !

z3dev: Done. Create poly2 geometry with minimal functionality. Updated all expansion functions to use poly2 for mearsureArea and arePointsInside.

z3dev:
@kaosat-dev please review the changes to the API again.

Summary:
- added 'corners' option, which accepts 'edge', 'chamfer', 'round'
- corrected useage of 'segments' options, which is now only used for creating round corners

In addition, poly2 geometry has been created. And, all expansion functions now use poly2 for measureArea and arePointsInside.

src/geometry/poly2/arePointsInside.js
```diff
@@ -0,0 +1,65 @@
+const measureArea = require('./measureArea')
```
kaosat-dev, line 1: great !

src/geometry/index.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
   geom2: require('./geom2'),
   geom3: require('./geom3'),
   path2: require('./path2'),
+  poly2: require('./poly2'),
```
kaosat-dev, line 5: groovy !

src/operations/expansions/expand.js
```diff
@@ -9,9 +9,10 @@ const expandPath2 = require('./expandPath2')
 /**
  * Expand the given object(s) using the given options (if any)
  * @param {Object} options - options for expand
- * @param {Number} options.delta=1 - delta (+/-) of expansion
- * @param {Integer} [options.segments] - number of segments when creating rounded corners, or zero for chamfer
- * @param {Object|Array} objects - the objects(s) to expand
+ * @param {Number} [options.delta=1] - delta (+/-) of expansion
+ * @param {String} [options.corners='round'] - type corner to create during of expansion; edge, chamfer, round
```
kaosat-dev: not perfect, but ok for now !

src/operations/expansions/expandGeom2.js
```diff
@@ -6,24 +6,31 @@ const offsetFromPoints = require('./offsetFromPoints')
...
+ * @param {String} [options.corners='round'] - type corner to create during of expansion; edge, chamfer, round
+ * @param {Integer} [options.segments=16] - number of segments when creating round corners
  * @param {geom2} geometry - the geometry to expand
  * @returns {geom2} expanded geometry
  */
 const expandGeom2 = (options, geometry) => {
   const defaults = {
     delta: 1,
-    segments: 0
+    corners: 'round',
```
kaosat-dev: hmm I was gonna say 'all ok', but did we not say that 'edge' should be the default ? (as that has no approximations etc)

kaosat-dev (changes requested):
Groovy @z3dev  almost there !
Only main thing remaining is the default type of corner (see my specific comment): I though we discussed 'edge' being the more logical default ? or is that wrong ?

src/operations/expansions/expandGeom2.js
z3dev: Yeah. I was going to make ‘edge’ the default, but V1 implementation is ‘round’.

I’m fine with edge but thought a short discussion would be best.

Thoughts?

z3dev: Done. It's 'edge' now.

src/operations/expansions/expandPath2.js
z3dev: Resolving. We can address this later if necessary.

z3dev:
@kaosat-dev please merge. :)

PR #185, opened 2019-08-19
Reorganization of Color

z3dev:
This pull request adds the color functions as a seperate module. The goals are:
- Full test coverage

color
- [X] code complete
- [X] tests complete, 81.25% coverage (due to error checks)
- [X] reveiw complete

Overall test coverage 83.87%.


z3dev:
This could be the last reorganization.

@kaosat-dev I believe this is the correct location of the color module. Please review.

src/color/color.js
```diff
@@ -0,0 +1,39 @@
...
+const colorNameToRgb = require('./colorNameToRgb')
+
+/**
+ * Apply the given color to the given objects.
+ * @param {Object} color - either an array or a hex string of color values
+ * @param {Object|Array} objects - the objects of which to color
+ * @returns {Object|Array} the same objects with an additional attribute 'color'
+ *
+ * @example
+ * let redSphere = color([1,0,0,1], sphere())
```
kaosat-dev: so the alpha for rgba is merged into the color array right ?
the only problem with that is for 'named/css' colors: they are rgb data, and without a seperate alpha parameter, we need to account for that somehow (hence some of the weirdness of the previous color op)

kaosat-dev: one possible solution, is to not do any 'magic' in here:
if you want a specific color, you (as a user) should create your colors using hexToRgba or colorNameToRgb etc
it has one extra perk : less code and conditionals here, and it means each functions does one thing only but one thing well :)

src/color/hexToRgb.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Converts a HTML5 color value (string) to RGB values.
+ * Conversion formula:
+ * - split the string; "#RRGGBB" into RGB components
+ * - convert the HEX value into RGB values
+ * @see the color input type of HTML5 forms
+ */
+const hexToRgb = (value) => {
```
kaosat-dev: I think this needs to work for rgbA hex values, (even more so if we remove the alpha parameter from the color() op), 
we already have such a thing here https://github.com/jscad/OpenJSCAD.org/blob/V2/packages/utils/img/src/colorConversions.js
This is one of the things I mentionned needing cleanup after we merge Openjscad.org & csg, so we can deal with it then ?

src/color/hslToRgb.js
```diff
@@ -1,21 +1,21 @@
 const hueToColorComponent = require('./hueToColorComponent')
 
 /**
- * Converts an HSL color value to RGB. Conversion formula
- * adapted from http://en.wikipedia.org/wiki/HSL_color_space.
+ * Converts an HSL color value to RGB.
  * Assumes h, s, and l are contained in the set [0, 1] and
  * returns r, g, and b in the set [0, 1].
  *
+ * @see http://en.wikipedia.org/wiki/HSL_color_space.
```
kaosat-dev, line 6: nice !

src/color/hslToRgb.js
```diff
@@ -1,21 +1,21 @@
...
  * @param   Number  h       The hue
  * @param   Number  s       The saturation
  * @param   Number  l       The lightness
  * @return  Array           The RGB representation
  */
 const hslToRgb = (h, s, l) => {
```
kaosat-dev: same remark as above, perhaps h, s, l , a to get a toRgba ?

src/color/hsvToRgb.js
```diff
@@ -1,19 +1,19 @@
...
- * @param   Number  s       The saturation
- * @param   Number  v       The value
- * @return  Array           The RGB representation
+ * @see http://en.wikipedia.org/wiki/HSV_color_space.
+ * @param Number h - hue
+ * @param Number s - saturation
+ * @param Number v - value
+ * @return Array the RGB representation
  */
 const hsvToRgb = (h, s, v) => {
```
kaosat-dev: same

src/color/index.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
kaosat-dev, line 1: nice & clean !

src/color/rgbToHex.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param  {Float} g - green value
+ * @param  {Float} b - blue value
+ */
+const rgbToHex = (r, g, b) => {
+  if (Array.isArray(r)) {
+    b = r[2]
+    g = r[1]
+    r = r[0]
+  }
+  let s = '#' +
```
kaosat-dev: es6 strings please :)

src/color/rgbToHsl.js
```diff
@@ -1,17 +1,17 @@
...
- * @return  Array           The HSL representation
+ * @see http://axonflux.com/handy-rgb-to-hsl-and-rgb-to-hsv-color-model-c
+ * @param Number r - red color value
+ * @param Number g - green color value
+ * @param Number b - blue color value
+ * @return Array the HSL representation
  */
 const rgbToHsl = (r, g, b) => {
-  if (r.length) {
+  if (Array.isArray(r)) {
```
kaosat-dev: way better !

src/color/rgbToHsv.js
```diff
@@ -1,20 +1,19 @@
...
- * adapted from http://en.wikipedia.org/wiki/HSV_color_space.
+ * Converts an RGB color value to HSV.
  * Assumes r, g, and b are contained in the set [0, 1] and
  * returns h, s, and v in the set [0, 1].
  *
- * @param   Number  r       The red color value
- * @param   Number  g       The green color value
- * @param   Number  b       The blue color value
- * @return  Array           The HSV representation
+ * @see http://en.wikipedia.org/wiki/HSV_color_space.
```
kaosat-dev, line 4: nice

kaosat-dev (changes requested):
clean and to the point @z3dev ! Only point to discuss is the alpha value related issues.

src/color/color.js
z3dev: I was thinking very similar. 

I think the color() function can just set color. And like you mentioned, the other functions can be used to translate between specifications or names.

But, I think most people will not provide the ALPHA.

We can have the color() function provide an ALPHA if not given (which the current code does)

OR

Like you said, let the viewer interpret colors if there’s no ALPHA.

Probably, we should do both.

src/color/rgbToHex.js
z3dev: I saw it... but the conversation to string is a wild one.

z3dev:
I will add some support for HEX ALPHA, RGBA and HSLA, as per web standards.

https://www.w3schools.com/cssref/css_colors_legal.asp

@kaosat-dev how about removing support for passing arrays? I don’t like the code that suppprts arrays, and actually had to fix most of the logic.

src/color/color.js
kaosat-dev: agreed, default to 1, but give the option to provide it (do  not give the option to set it in the color() function, but in the 'creator'/ converter function) :)

kaosat-dev:
@z3dev do you mean arrays passed to the various conversion functions ? I think that makes sense, users can use funct(...array) to have the same effect right ?
hmm then again, one of the foundations of V2 was the use of arrays (which is very practical in a ton of cases).
or switch to JUST arrays ?

src/color/hexToRgb.js
z3dev: Changed to accept an string notiation with RGB or RGBA hex values.

src/color/hslToRgb.js
z3dev: Changed to accept an array of HSL or HSLA values.

src/color/hsvToRgb.js
z3dev: Changed to accept an array of HSV or HSVA values.

z3dev:
>  or switch to JUST arrays ?

All conversion functions have been changed to accept an array of RGB/HSL/HSV values, only. I think this looks far better.

@kaosat-dev please review again

z3dev:
I created a new issue #187 to review these changes as part of the cleanup phase.

@kaosat-dev please merge if satisfied.

kaosat-dev:
well done @z3dev !

PR #189, opened 2019-10-15
V2 : Correcting use of 'size' parameter in primitives 

z3dev:
This pull request corrected the use / meaning of the 'size' parameter in the following primitives. Size now means full dimension, i.e. full width, length, depth, and height.
- cube
- cuboid
- rectangle
- roundedCuboid
- roundedRectangle
- square

All test suites that use these primitives have been adjusted as well.


z3dev:
As discussed, I have made changes to correct the use / meaning of 'size'.

@kaosat-dev please review and merge

src/primitives/rectangle.js
```diff
@@ -2,19 +2,20 @@ const vec2 = require('../math/vec2')
 
 const geom2 = require('../geometry/geom2')
 
-/** Construct an axis-aligned rectangle with four sides and four 90-degree angles.
+/**
+ * Construct an axis-aligned rectangle with four sides and four 90-degree angles.
  * @param {Object} [options] - options for construction
  * @param {Array} [options.center=[0,0]] - center of rectangle
- * @param {Array} [options.size=[1,1]] - size of rectangle, width and height
+ * @param {Array} [options.size=[2,2]] - full dimension of rectangle, width and length
```
kaosat-dev: I do not think it is necessary to specify 'full dimensions' here, that is the logical assumption when you say width and length I think :)

src/primitives/cuboid.js
```diff
@@ -12,15 +13,15 @@ const {geom3, poly3} = require('../geometry')
 const cuboid = (options) => {
   const defaults = {
     center: [0, 0, 0],
```
kaosat-dev, line 15: this is odd, I had not noticed the addition of center to all these ?

kaosat-dev (changes requested):
Looking good ! Just very minor tweak and merge is go !

src/primitives/cuboid.js
z3dev: Yup. Center is part of all primitives except cylinders.

It wasn't added. 'center' has always been part of the functionality, and came with the conversion of V1 to V2.

src/primitives/rectangle.js
z3dev: Done

z3dev:
This should be ready to go now.

@kaosat-dev please merge.

kaosat-dev (approved):

PR #190, opened 2019-10-16
V2 : Enable torus primitive

z3dev:
These changes enable the torus primitive, which requires the extrudeRotate() function.

- [X] code changes complete
- [X] test coverage complete
- [X] review complete

### DIFF
--- a/src/primitives/index.js
+++ b/src/primitives/index.js
@@ -18,5 +18,5 @@ module.exports = {
   sphere: require('./ellipsoid').sphere,
   square: require('./rectangle').square,
   star: require('./star'),
-//  torus: require('./torus')
+  torus: require('./torus')
 }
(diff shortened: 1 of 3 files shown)


z3dev:
By the way, I don’t think TORUS is a primitive because...
- it doesn’t use math or geometry modules
- it uses rotateExtrude(), which includes other modules, which includes other modules, etc.

If we want to keep this then it should be moved to extrusions, where it can live peacefully.

kaosat-dev:
hmm extrusions is a set of operations, so that would not be a good fit either
To me "primitive" does not describe how it is build, but more its geometric definition ?
'Shapes' would be a better fit ?

z3dev:
Good suggestions. I’ll add a new issue to follow this up.
For now, maybe we should just accept this.

kaosat-dev (approved):
All good!

kaosat-dev:
Indeed !

PR #191, opened 2019-10-17
V2 : Fix indexes

z3dev:
There was a lot going on during the renovation / the rewrite. :)

These changes correct the indexes of the various modules; adding missing functions, removing bad functions, and reordering the indexes.

All test pass, as this is only changes the index.js files.


z3dev:
Lots of missing functions in the indexes. This fixes those.

kaosat-dev (approved):
All good , thanks @z3dev !

PR #197, opened 2020-04-16
Delete duplicate code in fromSlices.js

ricochet1k:
fromSlices and addWalls are line-for-line duplicated in this file


z3dev:
@ricochet1k thanks for the changes.

this library has now been integrated into the OpenJSCAD.org repository, and now has a new package name as well; 'modeling'.

if you'd like to contribute these changes then please create a pull request against OpenJSCAD.org

by the way, we are looking for some early adopters to start using V2. if interested then here's a link to some more information.

https://openjscad.org/dokuwiki/doku.php?id=early_v2

